PR #24, opened 2026-01-19
Fix reconciler P0: ExecutionContext for mount error handling

roninjin10:
Adds ExecutionContext to control when components execute side effects, preventing mount errors in inactive phases/steps.

Changes:
- New `ExecutionContext.tsx` with `ExecutionProvider` and `useExecuteOnActive` hook
- Updated Phase/Step to wrap children with ExecutionProvider
- Updated all JJ/* and Git/* components to use `useExecuteOnActive` instead of `useMount`
- Fixed jsx-runtime typing conflicts

### DIFF
--- a/src/commands/db/current-view.ts
+++ b/src/commands/db/current-view.ts
@@ -1,37 +1,14 @@
 // Current execution view for database inspection
 
-interface Execution {
-  id: string
-  name?: string
-  status: string
-  file_path: string
-}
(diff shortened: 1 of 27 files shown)


roninjin10:
Superseded - main already has ExecutionContext implementation with a simpler pattern (checking execution.isActive in useMount). Closing as the core functionality is already merged.

PR #86, opened 2026-02-19
fix: tolerate codex rollout recorder stderr

SamuelLHuber:
this is something that just came up today for me, so not sure if others have the same issues. this local fix worked for me.

Problem: Codex can exit non-zero with only rollout recorder shutdown noise, which Smithers treated as fatal.

Fix: filter known rollout-recorder and cache TTL stderr lines; treat Codex non-zero as success when stderr is empty after filtering.

Assumptions: rollout-recorder and cache TTL errors are non-fatal if output schema is still produced.


roninjin10:
Superseded — rebased onto current main in fix/codex-rollout-recorder (original file was split in the agents refactor). Thank you @SamuelLHuber!

PR #87, opened 2026-02-23
smithers: resume --force and SIGINT cancellation

SamuelLHuber:
## Problem
When a smithers run process crashes or is interrupted, the run can remain stuck as `running` and cannot be resumed.

### Reproduction
1. Start a smithers run.
2. Kill the process (e.g. Ctrl+C).
3. Run `smithers list`.

### Observed Behavior
- After interrupting the process with Ctrl-C, `smithers list` still shows the run as status: `running`.
- `smithers resume` won’t attach because the run is not in a resumable state (it’s still marked `running`).

### Expected Behavior
- When the process exits, it is caught and the run is marked as failed/cancelled (not `running`).
- Either the run transitions to a resumable state after the process is stopped, or `resume` can attach to a run that is still marked `running`.

### Current Workaround
1. `smithers cancel ... --run-id <id>`
2. `smithers run ...` (starts a new run)

## Fix
- Handle SIGINT/SIGTERM in the CLI and pass an abort signal so interrupted runs are marked `cancelled`.
- Add `resume --force` to allow resuming a run still marked `running` (manual recovery path).

## Notes
- This keeps surface area minimal and avoids schema changes.

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -96,7 +96,7 @@ async function main() {
 
 Commands:
   run <workflow.tsx> [--input JSON] [--run-id ID] [--max-concurrency N]
-  resume <workflow.tsx> --run-id ID
+  resume <workflow.tsx> --run-id ID [--force]
   approve <workflow.tsx> --run-id ID --node-id ID [--iteration N] [--note TEXT] [--decided-by TEXT]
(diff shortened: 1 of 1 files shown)


SamuelLHuber:
one interesting idea maybe to add a heartbeat that sets a timestamp in the db so that runs can be clearly marked as stale/inactive even when status is running and could be marked as failed to allow resumption if e.g. 10 heartbeats are missed

PR #88, opened 2026-02-24
feat: add idle timeout for CLI agents

SamuelLHuber:
Adding an idle timeout so that if there is no agent response, the agent gets cancelled and can retry.

Issue being that the task timeout should be large so agents can fully run, do tests and verify their work. We want to trust the models.

But when they don't respond the long task timeout means it stalls the pipeline. therefore the idle timeout allows retrying that agent when it doesn't respond due to e.g. upstream api hang ups


roninjin10:
Good idea

PR #132, opened 2026-05-01
Honor non-retryable agent failures

cookesan:
## Summary
- preserve original agent generation failures so SmithersError code/details survive retry classification
- stop retrying agent failures marked failureRetryable=false or AGENT_CONFIG_INVALID
- keep catch, saga, and finally recovery paths from being mistaken for unhandled task failures

## Validation
- `pnpm -r typecheck`
- `pnpm lint`
- `pnpm test`

### DIFF
--- a/packages/components/src/components/WorktreeProps.ts
+++ b/packages/components/src/components/WorktreeProps.ts
@@ -1,6 +1,7 @@
 import type React from "react";
 
 export type WorktreeProps = {
+	key?: string;
 	id?: string;
 	path: string;
 	branch?: string;
--- a/packages/components/src/index.d.ts
+++ b/packages/components/src/index.d.ts
@@ -29,6 +29,7 @@ import * as _smithers_errors from '@smithers-orchestrator/errors';
 import * as zod_v4_core from 'zod/v4/core';
 
 type WorktreeProps$2 = {
+    key?: string;
     id?: string;
     path: string;
     branch?: string;
--- a/packages/engine/src/effect/workflow-bridge.js
+++ b/packages/engine/src/effect/workflow-bridge.js
@@ -72,8 +72,12 @@ function isRetryableBridgeTaskFailure(attempt) {
     if (meta?.failureRetryable === false) {
         return false;
     }
+    const errorCode = parseAttemptErrorCode(attempt?.errorJson);
+    if (errorCode === "AGENT_CONFIG_INVALID") {
+        return false;
+    }
     const kind = typeof meta?.kind === "string" ? meta.kind : null;
-    return !(kind !== "agent" && parseAttemptErrorCode(attempt?.errorJson) === "INVALID_OUTPUT");
+    return !(kind !== "agent" && errorCode === "INVALID_OUTPUT");
 }
 /**
  * @param {SmithersDb} adapter
--- a/packages/engine/src/engine.js
+++ b/packages/engine/src/engine.js
@@ -25,7 +25,7 @@ import { findVcsRoot } from "@smithers-orchestrator/vcs/find-root";
 import * as BunContext from "@effect/platform-bun/BunContext";
 import { eq, getTableName } from "drizzle-orm";
 import { getTableColumns } from "drizzle-orm/utils";
-import { Chunk, Duration, Effect, Fiber, Metric, Queue, Schedule } from "effect";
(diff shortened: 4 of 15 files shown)


roninjin10 (changes requested):
Review from Codex review agent; I am not fucory.

Findings:

1. P1 packages/scheduler/src/scheduleTasks.js:372
When `catchStatus.failed` is true, this returns before the `finally` block is traversed. A `TryCatchFinally` with try failed, catch failed, and finally still pending now produces no runnable cleanup work, and `makeWorkflowSession` will fail on the catch task immediately. That contradicts the component contract in `TryCatchFinally.js` that finally always runs after try success or catch failure. Please keep the catch failure suppressed long enough to schedule/run `finally`, then surface the catch failure after finally completes or fails.

2. P1 packages/scheduler/src/scheduleTasks.js:146 and packages/scheduler/src/scheduleTasks.js:351
Boundary inspection still treats any failed try/action child as the enclosing boundary failing, without considering whether an inner boundary already handled that failure. For example, an outer TCF whose try block contains an inner TCF with `innerTry=failed` and `innerCatch=finished` schedules the outer catch anyway. The new `includeContinuedFailures` path makes this especially likely for real TCF usage because try tasks are force-marked `continueOnFail`. `inspect` needs to model the nested boundary's recovery outcome, not just look through to its try/action failures.

3. P2 packages/scheduler/src/ScheduleResult.ts:15
`ScheduleResult` adds `failureRecoveryActive` / `failureRecoveryKeys`, but the committed public declaration entrypoint `packages/scheduler/src/index.d.ts` still omits both fields. `packages/scheduler/package.json` points the package and subpath `types` entries at that file, so consumers importing `ScheduleResult` or `scheduleTasks` will not see the runtime shape added here. Please update the public declarations or keep these fields out of the exported result type.

Verification I ran: `pnpm --filter @smithers-orchestrator/scheduler test` and `pnpm --filter @smithers-orchestrator/scheduler typecheck` passed on a separate PR worktree. The focused engine e2e test did not start locally because this machine's linked dependencies are missing `@effect/sql-sqlite-bun/SqliteClient`; the PR's GitHub `typecheck` and `faults` checks are passing.

cookesan:
Addressed the requested follow-ups on head `a73e399b`.

- `finally` now remains scheduled after a failed `catch`, and the `catch` failure is surfaced after cleanup completes.
- Boundary inspection now models nested saga and try/catch/finally recovery outcomes, so recovered inner failures do not trigger parent recovery.
- `ScheduleResult` public declarations now include `failureRecoveryActive` and `failureRecoveryKeys`.

I also rebased onto current `main` and kept the rebase cleanup isolated in a separate commit for the current type/build drift: `Worktree` JSX `key` typing, `gateway-react` provider `children` typing, and regenerated Gateway OpenAPI output from the generator.

Verified locally:
- `pnpm check:effect`
- `pnpm check:deps`
- `pnpm -r typecheck`
- `pnpm -r build`
- `pnpm --filter @smithers-orchestrator/scheduler test`
- `bun test packages/engine/tests/failure-retryable-false.e2e.test.jsx`
- `pnpm --filter @smithers-orchestrator/e2e test:faults`

Fresh GitHub Actions runs for this head are currently `action_required` with no jobs started, so they need maintainer approval before remote CI can execute.

roninjin10 (approved):
Review from Codex review agent; I am not fucory.

No blocking findings on the current head (`a73e399b`). The follow-up changes address my previous scheduler concerns: `finally` is scheduled before surfacing a failed catch, nested recovered try/catch/finally boundaries no longer trigger parent recovery, and the public `ScheduleResult` declaration now includes the recovery fields.

Verified locally:
- `pnpm --filter @smithers-orchestrator/scheduler test`
- `pnpm --filter @smithers-orchestrator/scheduler typecheck`
- `bun test packages/engine/tests/failure-retryable-false.e2e.test.jsx`
- `pnpm --filter @smithers-orchestrator/components typecheck`
- `pnpm --filter @smithers-orchestrator/gateway-react typecheck`
- `pnpm --filter @smithers-orchestrator/gateway typecheck`
- `pnpm --filter @smithers-orchestrator/gateway check:openapi`
- parsed `packages/gateway/openapi.yaml`

Remote `typecheck` and `faults` checks are passing.

PR #133, opened 2026-05-04
Harden gateway client contracts

cookesan:
## Summary
- add contract tests for gateway-client HTTP RPC, WebSocket RPC, handshake, stream filtering, and malformed-frame paths
- harden gateway-client error handling for invalid HTTP/WebSocket frames, failed sends, closed sockets, aborted opens, and rejected handshakes
- add typed helpers for hijack, rewind, cron RPCs, and DevTools streams, then expose the expanded write surface through gateway-react actions
- make useGatewayRpc dependency handling lint-clean while preserving explicit dependency control

## Validation
- pnpm --filter @smithers-orchestrator/gateway-client test
- pnpm --filter @smithers-orchestrator/gateway-client typecheck
- pnpm --filter @smithers-orchestrator/gateway-react test
- pnpm --filter @smithers-orchestrator/gateway-react typecheck
- pnpm lint
- pnpm typecheck
- pnpm test

### DIFF
--- a/packages/gateway-client/src/SmithersGatewayClient.ts
+++ b/packages/gateway-client/src/SmithersGatewayClient.ts
@@ -15,6 +15,13 @@ type StreamRunEventPayload = {
   payload?: unknown;
 };
 
+type StreamDevToolsEventPayload = {
+  streamId?: string;
+  runId?: string;
+  event?: unknown;
+  error?: unknown;
+};
+
 declare global {
   var __SMITHERS_GATEWAY_UI__: GatewayUiBootConfig | undefined;
 }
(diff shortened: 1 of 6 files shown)


roninjin10:
Nice!

roninjin10:
> 🤖 **This is an automated review from OpenAI Codex (GPT-5)**, requested by @roninjin10 / @fucory. Not a human review. Treat as a second opinion alongside human review, not a replacement.

## Summary

This PR moves the client in the right direction: response validation is stricter, failed sends no longer leak pending requests, and the new helper surface is mostly covered. I would not merge as-is, though, because several failure paths can still leave `connect()` / `requestRaw()` promises hanging forever, which is exactly the class of contract-hardening bug this PR is trying to eliminate.

## Strengths

- HTTP RPC now distinguishes malformed successful gateway responses from plain HTTP failures, which is a real improvement over trusting `response.json()` blindly.
- WebSocket send failures are handled correctly by removing the pending request before rejecting.
- Rejected WebSocket handshakes now close the socket, avoiding an obvious resource leak.
- The new tests cover the happy-path contract, RPC error preservation, malformed HTTP responses, failed WebSocket sends, rejected handshakes, and stream-id filtering.

## Issues

1. **P2 — `packages/gateway-client/src/SmithersGatewayClient.ts:177` — `connect()` can hang forever if the socket closes before `open`.**

   The open promise listens for `"open"` and `"error"`, but not `"close"`. A WebSocket can close during connection setup without producing a useful `"error"` event in all environments. In that case, `await new Promise(...)` never resolves or rejects.

   Suggested fix: add an `onClose` listener during the open phase, clean it up with the others, and reject with a clear `"Gateway WebSocket closed before open"` error. Add a test where the fake socket dispatches `"close"` before `"open"`.

2. **P2 — `packages/gateway-client/src/SmithersGatewayClient.ts:184` / `:203` — abort only applies until the socket opens, not through the handshake.**

   The abort listener is removed as soon as `"open"` fires, but the connection is not usable until `connection.requestRaw("connect", ...)` resolves. If the signal aborts after open but before the handshake response, the returned `connect()` promise remains pending and the socket stays open.

   Suggested fix: keep abort handling active through the handshake, or race the handshake promise against the abort signal and call `connection.close()` on abort. This also matters for `streamDevTools()` and existing stream helpers because they can leak during setup before `connection.events(signal)` starts consuming.

3. **P2 — `packages/gateway-client/src/SmithersGatewayConnection.ts:178` / `:213` — malformed WebSocket frames do not reject pending RPCs.**

   Invalid JSON and invalid non-response frames are pushed into the event queue, but `requestRaw()` callers usually are not consuming `events()`. During `connect()`, for example, a malformed server frame will queue an error while the pending handshake promise waits forever.

   Suggested fix: treat protocol-invalid inbound frames as fatal for the connection: reject all pending requests and close the socket, or at minimum reject pending requests when the invalid frame cannot be associated with a specific request. Add tests for malformed frames during `client.connect()` and during a plain `requestRaw()` with no event iterator.

4. **P2 — `packages/gateway-client/src/SmithersGatewayClient.ts:157` — HTTP RPC can succeed on a non-2xx response if the body is `{ ok: true }`.**

   After validating the frame, the code returns `frame.payload` whenever `frame.ok` is true, even if `response.ok` is false. That means a `500`/`502` response with a syntactically valid success frame is treated as success, which violates the HTTP contract and can mask proxy/gateway failures.

   Suggested fix: after validation, if `!response.ok && frame.ok`, throw `gatewayHttpError(method, response.status)` or an invalid-response error. Add a regression test for a `500` response containing an `ok: true` frame.

5. **P3 — `packages/gateway-react/tests/gateway-react.test.ts:1` — the `useGatewayRpc` dependency rewrite has no behavioral test coverage.**

   The PR changes `useGatewayRpc` substantially, but the new React tests only use `renderToString`, which does not run effects. So the risky part of the React change, dependency-controlled fetching, is not tested at all.

   Suggested fix: add a client-side hook test using `act()` with a renderer that runs effects. Cover default params-triggered refetch, custom `deps`, `enabled` toggling, and manual `refetch()` using the latest params.

## Nits

- `isObject()` and `isGatewayResponseFrame()` are duplicated between `SmithersGatewayClient.ts` and `SmithersGatewayConnection.ts`. Consider moving the frame validators into a small shared internal module so the HTTP and WebSocket contracts cannot drift.
- The tests assert `connection.pending.size`, which codifies an internal implementation detail. Prefer testing observable behavior and making `pending` private in a follow-up.

## Verdict

**PASS_WITH_FIXES** — the direction is good, but the remaining WebSocket hang/leak paths should be fixed before merging because they undermine the hardening goal.

— Codex (GPT-5, via OpenAI Codex CLI · `model_reasoning_effort=high`)

PR #142, opened 2026-05-20
Add eval suites for workflow regression

cookesan:
Summary
- Add `smithers eval` for JSON and JSONL workflow regression suites.
- Support dry-run planning, run labels, bounded case concurrency, status checks, output assertions, error substring checks, and persisted JSON reports.
- Document eval suites in the CLI catalog, quickstart guide, recipes, and README.

Validation
- `pnpm --filter @smithers-orchestrator/cli typecheck`
- `pnpm --filter @smithers-orchestrator/cli test`
- `pnpm typecheck`

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,10 +78,25 @@ Each task output is validated against its Zod schema and persisted to SQLite. If
 ```bash
 smithers up workflow.tsx --input '{"description": "Fix bug"}'
 smithers up workflow.tsx --run-id abc123 --resume true
+smithers eval workflow.tsx --cases evals/smoke.jsonl --suite smoke
 smithers ps
 smithers workflow list
 smithers approve abc123 --node review
 ```
+
+## Eval suites
+
+Run repeatable workflow regressions from JSON or JSONL cases:
+
+```jsonl
+{"id":"happy-path","input":{"description":"Fix bug"},"expected":{"status":"finished"}}
+```
+
+```bash
+smithers eval workflow.tsx --cases evals/smoke.jsonl --suite smoke --force
+```
+
+Reports are written to `.smithers/evals/<suite>.json` and the command exits non-zero when any case fails.
 
 ## Hot Reload
 
--- a/apps/cli/src/eval-suite.js
+++ b/apps/cli/src/eval-suite.js
@@ -5,6 +5,7 @@ import { SmithersError } from "@smithers-orchestrator/errors";
 
 export const EVAL_CASE_STATUSES = [
     "finished",
+    "continued",
     "failed",
     "cancelled",
     "waiting-approval",
@@ -13,6 +14,7 @@ export const EVAL_CASE_STATUSES = [
 ];
 
 const RUN_ID_MAX_LENGTH = 64;
+const EVAL_EXPECTED_KEYS = new Set(["status", "output", "outputContains", "errorContains"]);
 
 /**
  * @param {unknown} value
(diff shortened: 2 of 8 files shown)


roninjin10:

---

**Findings**

1. **Duplicate normalized case IDs can collide into the same run and corrupt scoring**  
   `apps/cli/src/eval-suite.js:150` normalizes `id`/`name` into a slug, and `apps/cli/src/eval-suite.js:268` derives run IDs from that slug, but no uniqueness check is performed. Cases like `"A B"` and `"a-b"`, repeated IDs, or long IDs that truncate/hash the same way can produce identical `caseId`/`runId`. `apps/cli/src/index.js:2752` then runs both cases against the same persisted run ID, so results can reuse or overwrite prior state and report misleading passes/failures. The loader should reject duplicate normalized case IDs/run IDs before execution.

2. **Re-running the same eval run ID can reuse stale workflow state instead of executing a fresh regression**  
   `apps/cli/src/index.js:2752` calls `runWorkflow` directly with deterministic eval run IDs, but unlike `smithers up`, the eval command does not check whether those run IDs already exist or clear their persisted nodes/outputs. With `--run-label ci --force`, the report is overwritten, but the run IDs are identical; existing finished attempts/outputs can be treated as already completed by the engine. This creates false positives after workflow changes. Eval execution should either require non-existing run IDs, generate unique run IDs unless explicitly resuming, or reset/fork run state before execution.

3. **Unknown `expected` assertion keys are silently ignored, causing false passes**  
   `apps/cli/src/eval-suite.js:87` accepts and preserves arbitrary keys in `expected`, but `apps/cli/src/eval-suite.js:308`, `apps/cli/src/eval-suite.js:316`, and `apps/cli/src/eval-suite.js:324` only evaluate `output`, `outputContains`, and `errorContains`. A typo like `outputsContains` falls back to status-only scoring and can pass while not testing the intended condition. `normalizeExpected` should reject unknown keys.

4. **`errorContains` does not work for normal failed workflow results**  
   `apps/cli/src/index.js:2773` passes the `runWorkflow` result into `evaluateEvalCaseResult`; failed runs commonly return a structured `error` object. `apps/cli/src/eval-suite.js:325` converts that object with `String(result.error)`, yielding `"[object Object]"`, so `expected.errorContains: "boom"` will fail even when the actual error message contains `"boom"`. The evaluator should extract `error.message`, serialize structured errors, or use the same error formatting as reports/CLI output.

5. **The eval status enum omits a real engine terminal status**  
   `apps/cli/src/eval-suite.js:6` allows `finished`, `failed`, `cancelled`, and waiting states, but the engine can return `continued` for continue-as-new workflows. Such workflows cannot be represented as expected outcomes, despite being valid workflow behavior. This is a regression-coverage gap for long-running workflow suites.

6. **Documented output shape is not stable across workflows**  
   The guide says output assertions match "the persisted output snapshot, keyed by output name" (`docs/guides/evals-quickstart.mdx:42`), but the command only loads the persisted snapshot when `result.output` is `undefined` (`apps/cli/src/index.js:2769`). For workflows whose schema has an `output` table, `runWorkflow` returns the raw rows array, while other workflows get the keyed snapshot from `loadOutputs`. Eval cases will need different assertion shapes for otherwise similar workflows. Prefer always scoring against a normalized snapshot, or document the exact shape.

**Verdict: REQUEST_CHANGES**

The feature direction is useful, but the current harness has scoring-soundness problems: assertion typos can pass silently, failed-run error matching is broken, deterministic run IDs can collide or reuse stale persisted state, and continue-as-new workflows are not representable. I would fix those before merging, then add tests specifically for duplicate IDs, rerunning with the same `--run-label`, failed workflows with `errorContains`, unknown expected keys, and `continued` status handling.

cookesan:
Addressed in 0394be2b. Eval loading now rejects duplicate normalized case ids/run ids and unknown expected assertion keys, failed-run matching formats structured errors correctly, continued is accepted as a terminal expectation, reruns refuse existing eval run ids before execution, and scoring uses a normalized output snapshot consistently. Local validation: bun test apps/cli/tests/eval-suite.test.js, git diff --check. Hosted typecheck and faults are green.

roninjin10 (approved):
Cool feature happy to merge this. I think in future we need to think about how we could do evals for individual steps in the workflow though

roninjin10:
Reviewed the merged eval suites PR and found two small follow-up fixes: make `outputContains` array matching behave like a real recursive partial match instead of a prefix match, and add the missing `continued` status to the evals quickstart docs. We’re fixing both in a follow-up.

PR #143, opened 2026-05-20
Expand hosted controls and CLI hardening

cookesan:
## Summary
- add hosted control-plane identity provider records and usage-limit checks, including audit export coverage
- harden CLI preflight behavior for unknown commands, resume-claim pairing, stdin JSON input, and command docs coverage
- make sandbox cleanup remove shipped request workspaces while preserving collected result bundles

## Validation
- pnpm test
- pnpm -r typecheck
- pnpm -r build
- pnpm lint (8 warnings, 0 errors)
- git diff --check
- GitHub CI: test, typecheck, faults passed


roninjin10:

---

## P0

None found.

## P1

- `packages/sandbox/src/effect/process-runner.js:17` leaks the Gateway/worker process environment into sandbox command execution. `spawnSandboxCommand(..., { env: process.env })` means bubblewrap and macOS `sandbox-exec` commands inherit all ambient secrets, tokens, cloud creds, DB URLs, etc. That defeats the "untrusted sandbox" boundary and contradicts the secret-handling guidance added in the docs. Use a minimal allowlisted env, ideally `--clearenv` for bubblewrap, and only pass explicit `config.env` entries after validation/redaction.

- `packages/sandbox/src/effect/process-runner.js:101` and `packages/sandbox/src/SandboxHandle.ts:7` only carry/enforce `image` and `allowNetwork`; `env`, `ports`, `volumes`, `memoryLimit`, `cpuLimit`, and workspace controls are silently ignored by the live command runners. This is a security/operational footgun: callers can configure limits believing they protect the host, but Docker/bubblewrap run without them. At minimum Docker should map memory/CPU/env/volumes/ports explicitly, and unsupported controls should fail closed rather than being ignored.

- `packages/server/src/gatewayUi/defaultConsole.js:5` persists the bearer token in `localStorage`, and `packages/server/src/gateway.js:197` exposes the built-in operator console without any server-side UI auth gate. Any script served from the Gateway origin, including custom/workflow UIs or an XSS in the console, can read `smithers.gateway.token` and submit approval decisions. For hosted control surfaces, this should use an HttpOnly/session-backed auth path or memory-only token storage, and the console route itself should be protected when bearer auth is configured.

- `packages/db/tests/db-concurrent.test.js:196` enables the previously skipped cross-adapter transaction serialization test, but this diff does not include a corresponding fix to the per-adapter transaction queue described in the test comment. The only production DB changes are cache upsert/schema work, so this looks like it will still reproduce `cannot start a transaction within a transaction` and fail `pnpm test` in the new CI job.

## P2

- `packages/engine/src/engine.js:2758` still includes `nodeId`, `iteration`, and `outputTableName` in `cacheBase` even when `cachePolicy.key` is provided, while `packages/engine/src/cache-policy.js:18` documents/builds a separate `taskKey`. That means `key` does not actually override the task-ID-derived identity, and `scope: "global"` only shares when distinct workflows happen to use the same task id/output table. Add tests where two workflows use different task ids with the same cache key.

- `packages/engine/src/hot/watch.js:116` adds a recursive stat poll of the entire watched tree every 50-250ms. On real repositories this can become a constant CPU and filesystem load even while idle, especially because every interval walks all non-ignored files. This needs a much larger/backoff interval, a file count guard, or a platform-specific fallback only when native watch events are known to be unreliable.

- `packages/control-plane/src/index.js:134` defines `_smithers_cp_secret_refs` with `PRIMARY KEY (org_id, project_id, name)` while `project_id` is nullable. In SQLite, nullable columns in a composite primary key can allow multiple org-wide rows for the same `(org_id, NULL, name)` under direct writes or races. The usage-limit table already uses a non-null `project_key`; secret refs should use the same pattern or a partial unique index for `project_id IS NULL`.

- `packages/sandbox/src/effect/process-runner.js:101` mounts Docker `/workspace` as read-only but then runs arbitrary `command` with `/result` writable and no validation of expected output paths beyond later bundle collection. That may be okay for trusted commands, but for untrusted commands the result directory becomes the only write boundary. The runner should explicitly document/enforce that command output must be under `/result`, and reject symlink/path traversal on collection for all runtime-produced files.

## P3

- `packages/control-plane/package.json:3` introduces `@smithers-orchestrator/control-plane` at version `0.20.3` while the monorepo and facade package are `0.20.4`. That is likely to confuse release/publish automation and consumers expecting aligned workspace versions.

- `packages/control-plane/src/index.js:25` uses `SLUG_RE = /^[a-z0-9][a-z0-9-]{0,62}[a-z0-9]$/`, which rejects one-character slugs despite otherwise looking like a standard slug validator. If one-character org/project slugs are not intentionally forbidden, adjust the regex and add a test.

## Additional findings from `codex review --base main` (default review pass)

- **[P2] Serialize stdin annotations for detached runs** — `apps/cli/src/index.js:1514-1516`. When `smithers up ... --detach --annotations -` is used, this path parses annotations from the parent process' stdin but still forwards the literal `-` to the background child. The child no longer has the original stdin, so it re-reads empty input and fails with `INVALID_JSON` (or otherwise loses the annotations); this needs the same serialization treatment as `--input -`.

- **[P2] Allow documented `--mcp --surface` form** — `apps/cli/src/index.js:5399-5402`. For documented invocations like `smithers --mcp --surface raw`, this pre-validation runs before MCP argument parsing, and `findFirstPositionalIndex` does not treat `--surface` as a value-taking flag. The value (`raw`, `semantic`, or `both`) is therefore interpreted as an unknown command and the MCP server exits before starting; only the undocumented `--surface=raw` form still works.

---

**VERDICT: REQUEST_CHANGES**

The PR adds useful surfaces, but the sandbox execution changes and hosted console introduce security-sensitive regressions: ambient secrets are passed into local sandbox commands, several advertised sandbox controls are ignored, and the built-in console stores operator bearer tokens in origin-readable localStorage. I would also fix the likely failing unskipped DB concurrency test, tighten the cache-key semantics, and restore the documented `--detach --annotations -` and `--mcp --surface <value>` CLI flows before merging.

---

<sub>Generated by `codex review --base main -c model_reasoning_effort=high` + `codex exec -c model_reasoning_effort=high` (OpenAI Codex CLI v0.133.0, model `gpt-5.5`). Posted via automation on @roninjin10's account; content is not a human review.</sub>

cookesan:
Addressed in 1460d226. Sandbox runners now use a minimal runner env, pass explicit controls, fail closed when a runtime cannot enforce a control, and reject symlinked bundle outputs; the built-in console no longer persists tokens and is gated by Gateway bearer auth when auth is configured; cache keys now honor cachePolicy.key across task ids; hot watch polling backs off with a file-count guard; secret refs use a non-null project_key with migration coverage; and the detached --annotations - plus --mcp --surface raw CLI paths are covered. Local validation: pnpm test, touched package typechecks, git diff --check, and architecture budget all pass. CI is running on the pushed commit now.

cookesan:
Follow-up in 109cc1a8: the CI test job was stopping at the architecture budget because apps/cli/src/index.js had grown past the command-registry limit. I moved the generic argv helpers into apps/cli/src/argv-utils.js, keeping the CLI behavior unchanged while bringing index.js back under budget. Local validation: pnpm test, pnpm --filter @smithers-orchestrator/cli typecheck, bun test apps/cli/tests/cli-flag-validation.test.js apps/cli/tests/agents-ts-codegen.test.js, node scripts/check-architecture-budget.mjs, and git diff --check all pass. CI is running on the pushed commit now.

roninjin10:
gonna break this up into smaller more atomic prs

roninjin10:
Closing in favor of 7 atomic per-domain PRs that preserve your authorship on every commit:

- #148 — `ci: add test job + workflow catalog docs + agent test timing fix`
- #149 — `db: split SqlMessageStorage into lowercase module + cache upsert`
- #150 — `engine: extract cache policy + harden hot watch + jumpToFrame`
- #151 — `sandbox: add process-runner + safety boundaries + shipped-bundle cleanup`
- #152 — `server: extract gateway UI into focused modules + default console`
- #153 — `cli: preflight, json-args, argv-utils + architecture budget script`
- #154 — `control-plane: add hosted package + rebrand + production hardening guide`

Each one is independently reviewable and mergeable in any order.

Two conflicts vs current `main` (from #142 eval-suites and `d03daa7d` raw-json fix landing after this PR opened) were resolved in the splits:
- `apps/cli/src/index.js` `KNOWN_COMMANDS` — kept your programmatic `...cliCommands.keys()` form (auto-includes `eval`).
- `README.md` — kept both new sections (Gateway Console + Eval suites).

One intentional deviation: in #153 the `cli/index.js` architecture budget is `5700` instead of your `5480` target, to fit the `eval` command that landed in #142. Flagged in the PR body.

Thanks for the work — splitting just made it easier to land piece by piece.

PR #144, opened 2026-05-21
Harden runtime durability, sandboxing, and operator console

cookesan:
## Summary
- enforce run-owned SQLite foreign keys, legacy FK rebuild migrations, and cross-adapter transaction serialization
- make Bubblewrap sandbox execution real, cleanup-backed, and environment-isolated
- implement cache policy ttl/scope behavior and add a default operator console at /console
- add a standard CI test job with a pinned Bun version

## Verification
- pnpm test
- pnpm -r typecheck
- pnpm -r build
- pnpm lint (passes with 8 existing warnings)
- git diff --check

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,3 +32,27 @@ jobs:
       - run: pnpm -r typecheck
 
       - run: pnpm --filter ./.smithers typecheck
+
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: pnpm/action-setup@v4
+        with:
+          version: 10.10.0
+          run_install: false
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: pnpm
+
+      - uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: 1.3.13
+
+      - run: pnpm install --frozen-lockfile
+
+      - run: pnpm test
--- a/apps/cli/src/agent-commands/regenerateAgentsTsIfPresent.js
+++ b/apps/cli/src/agent-commands/regenerateAgentsTsIfPresent.js
@@ -1,6 +1,6 @@
 import { existsSync, readFileSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
-import { generateAgentsTs } from "../agent-detection.js";
+import { extractGeneratedDetectionProviderIds, generateAgentsTs } from "../agent-detection.js";
 
 /**
  * If the current working directory contains a `.smithers/agents.ts` that was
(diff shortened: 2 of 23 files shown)


roninjin10:

---

**Verdict: REQUEST_CHANGES**

The patch introduces functional regressions in TTL cache refresh, generated agents preservation for account labels that collide with detector IDs, and the macOS sandbox fallback's writable temp handling. These issues can break advertised behavior and should be fixed before considering the patch correct.

### Full review comments

- **[P2] Replace expired cache entries instead of leaving stale rows** — `packages/engine/src/engine.js:2813-2823`
  When `ttlMs` expires, this branch treats the row as a miss but leaves the old `_smithers_cache` row in place. The later completion path still calls `insertCache`, which is implemented as insert-ignore, so the refreshed payload and `createdAtMs` are never written while the stale key exists; after the first expiration, every subsequent run for that key will re-execute forever instead of caching for another TTL window.

- **[P2] Avoid preserving account labels as detected agents** — `apps/cli/src/agent-detection.js:176-179`
  For generated accounts-only files, an account label like `codex` or `claude` is rendered as a provider key with the same name as a detector. This extractor then treats that account-backed provider as detection-derived, so the next regeneration can force the detection shape and emit scaffolded local imports/providers even when no CLI was detected, contradicting the comment that account labels are ignored and potentially breaking projects that only have accounts configured.

- **[P2] Permit a writable temp directory in the macOS sandbox** — `packages/sandbox/src/effect/socket-runner.js:101-102`
  On the macOS `sandbox-exec` fallback, the process environment still sets `HOME`/`TMPDIR` to `/tmp`, but the profile only permits writes under the request and result directories. Any sandboxed command or tool that writes temporary files or caches through `$TMPDIR`/`$HOME` (for example `mktemp`, Bun/Node tooling, or `smithers up`) will fail with a sandbox denial even though the Linux bwrap path creates a writable `/tmp`.

---

_Generated by Codex CLI 0.133.0 (`gpt-5.5`, reasoning effort: high). Verdict inferred by the automation harness from the three P2 functional-regression findings._

cookesan:
Addressed in 79b2bf42. Expired TTL cache entries now update the existing cache row after refresh; generated agents preservation now only carries forward detector-owned provider initializers, so account labels that collide with provider ids stay account-backed; and the macOS sandbox fallback now runs with a sandbox-owned writable HOME/TMPDIR covered by the sandbox profile. I also fixed the hot watcher fallback exposed by the full test run so missed native events are picked up by a guarded poll without making startup depend on the initial scan. Local validation: pnpm test, focused regression tests, touched package typechecks, and git diff --check all pass. CI is running on the pushed commit now.

roninjin10 (dismissed):
We are going to need to add the concept of migrations to smithers for stuff like this

roninjin10:
I'm going to add it in a pr first and then rebase this pr and merge after

PR #160, opened 2026-05-25
Add cleanup followups

lalalune:
## What changed

- Splits the non-demo dirty worktree changes into a separate cleanup PR.
- Includes CLI/devtools error handling and JSON stdout contract updates.
- Includes agent diagnostics, SQL message storage streaming, engine deferred/heartbeat/workflow command coverage, driver child process cleanup, and time-travel fork recovery changes.
- Includes the associated docs/reference and Smithers ticket updates.

## Validation

- `PATH=/Users/shawwalters/.bun/bin:$PATH bun test --timeout=120000 --max-concurrency=1 apps/cli/tests/cli-devtools-process.test.js apps/cli/tests/docs-examples-smoke.test.js apps/cli/tests/init.e2e.test.js apps/cli/tests/json-stdout-contract.test.js packages/agents/tests/agent-diagnostics.test.js packages/db/tests/sql-message-storage-stream.test.js packages/engine/tests/deferred-contract.test.js packages/engine/tests/heartbeat-bounds.test.jsx packages/engine/tests/workflow-command.e2e.test.js packages/time-travel/tests/fork-recovery.test.js`

Result: 81 pass, 1 expected skip, 0 fail.

### DIFF
--- a/.smithers/tickets/0003-6larp-assessment.md
+++ b/.smithers/tickets/0003-6larp-assessment.md
@@ -1,0 +1,30 @@
+# 6LARP Assessment TODO
+
+Objective: critically evaluate whether shipped code is real or performative, then fix proven issues from most complicated to simplest.
+
+## Findings
+
+- [x] `packages/time-travel/src/fork/forkRunEffect.js` writes the child snapshot/run before the branch row without an enclosing transaction. If branch persistence fails, tests document that partial child state can be left behind. Fixed by making fork creation atomic and updating the mid-creation failure test to prove rollback.
+- [x] `packages/db/src/sql-message-storage.js` exposes `SqlClient.Connection.executeStream` but implements it as `Effect.dieMessage("executeStream not implemented")`. This was a real stub on an exported internal SQL client surface. Replaced with a functional stream and added a test that executes it.
+- [x] Root dependency-boundary check failed because `apps/smithers-demo` imported `react` and `react-dom/client` without declaring `react` / `react-dom` in its own package manifest. Added direct dependencies to the app manifest.
+- [x] `packages/agents/tests/agent-diagnostics.test.js` had an async test that only asserted a promise was returned, then left the diagnostics work running in the background. Fixed it to await the diagnostics report and assert its contents.
+- [x] `packages/driver/src/child-process.js` could let a stale idle timer kill a process after stdout activity reset the watchdog. Added a generation guard and a Bun/macOS stdout-delivery grace for long CLI idle windows; targeted timeout tests now prove stdout activity keeps the process alive while hard timeouts still win.
+- [x] `packages/engine` E2E tests exercised real CLI subprocesses and durable workflow paths but relied on Bun's 5s default test timeout. In a full package/workspace run they timed out despite passing with a realistic timeout budget. Made the package script explicit with a 60s timeout.
+- [x] `packages/engine/tests/loop-until-reeval.test.jsx` passed alone but timed out inside the package/workspace run, which exposed non-isolated engine tests sharing process/runtime state under Bun's file concurrency. Made the engine package test script serial (`--max-concurrency=1`) so these integration tests execute in the same isolation mode they require.
+- [x] `packages/engine/tests/deferred-contract.test.js` asserted a 120ms timer would still be waiting after a full workflow start, which is a wall-clock race under load. Increased the timer margin so the test validates pause/resume behavior instead of scheduler speed.
+- [x] `packages/smithers/tests/e2e-helpers.js` spawned CLI E2E subprocesses without a timeout, so a hung CLI could block until Bun killed a dangling child. Added a bounded default timeout and stable SIGTERM exit mapping.
+- [x] `packages/engine/tests/heartbeat-bounds.test.jsx` used a 30s explicit timeout for real 1MB heartbeat persistence and skipped cleanup on failure. Raised the timeout to the package budget and wrapped DB cleanup in `finally`.
+- [ ] Skipped tests in `packages/time-travel/tests/rewindLock-concurrent.test.ts`, `packages/time-travel/tests/fork-recovery.test.js`, `packages/sandbox/tests/sandbox-isolation.test.js`, and the fault-injection suite are honest documented environment/feature gaps, not fake-green assertions. Keep them flagged as unproven unless their prerequisites are implemented.
+- [ ] Broad 6LARP scan found many `return null` / `return []` paths in React marker components, parsers, optional resolvers, and agent event decoders. These are not automatically stubs, but only targeted tests prove specific behavior.
+
+## Verification TODO
+
+- [x] Run targeted DB tests covering `executeStream`.
+- [x] Run targeted time-travel fork recovery tests covering rollback.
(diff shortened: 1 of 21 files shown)


roninjin10 (commented):
Codex-only review: I am OpenAI Codex running through the `gh` CLI, not fucory. fucory will review immediately after this.

No blocking findings. I focused on the non-cosmetic behavior in this cleanup PR: `executeStream` now exercises the SqlClient stream path, fork creation is wrapped in one transaction, the diagnostics test now awaits the report, and the child-process watchdog tests still cover timeout/cancellation behavior.

Validation I ran in a temporary worktree:

- `pnpm install --frozen-lockfile --ignore-scripts`
- `bun test packages/db/tests/sql-message-storage-stream.test.js` - pass
- `bun test packages/time-travel/tests/fork-recovery.test.js` - pass, 5 pass / 1 expected skip
- `bun test packages/agents/tests/agent-diagnostics.test.js` - pass, 25 tests
- `bun test packages/driver/tests/spawnCaptureEffect.test.js` - pass, 26 tests
- `git diff --check origin/main...refs/remotes/origin/pr/160` - pass

roninjin10:
### Storyline walkthrough from Codex

I am Codex, not fucory. fucory will review immediately after this.

This PR is a cleanup and hardening pass. It is not one single feature. It reads like a sweep through places where the code already mostly worked, but some exported behavior, failure recovery, or tests were weaker than they looked.

The easiest way to understand it is as six small cleanups that make the system more honest and durable.

## 1. A real SQL stream replaces a stub

`packages/db/src/sql-message-storage.js` exposed an Effect SQL connection with `executeStream`, but the method was still a stub.

The meaningful diff is:

```diff
- executeStream: () => Effect.dieMessage("executeStream not implemented"),
+ executeStream: (statement, params, transformRows) =>
+   Stream.fromIterableEffect(execute(statement, params, transformRows)),
```

So callers that go through the Effect `SqlClient` stream path now get actual SQLite rows instead of a runtime death.

The new test proves the path end to end:

```js
const rows = await storage.withConnection((connection) =>
  Stream.runCollect(
    connection.executeStream(
      "SELECT id, name FROM stream_items ORDER BY id",
      [],
      undefined,
    ),
  ).pipe(Effect.map(Chunk.toReadonlyArray)),
);

expect(rows).toEqual([
  { id: 1, name: "alpha" },
  { id: 2, name: "beta" },
]);
```

## 2. Fork creation becomes atomic

`packages/time-travel/src/fork/forkRunEffect.js` creates a child run from a parent snapshot. Before this PR, the fork flow wrote pieces one by one: snapshot first, run metadata second, branch row last. If the branch write failed, partial child state could be left behind.

The old shape was effectively:

```text
insert child snapshot
insert child run
insert branch relationship
```

The new shape is:

```js
yield* adapter.withTransactionEffect("fork run", Effect.gen(function* () {
  yield* insertChildSnapshot;
  if (childRun) {
    yield* adapter.insertRun(childRun);
  }
  yield* insertBranchRelationship;
}));
```

That changes the contract to all-or-nothing:

```text
snapshot + run + branch all commit
or
snapshot + run + branch all roll back
```

The recovery test now checks that a simulated branch insert failure does not leave an orphan child run or child snapshot behind.

## 3. Child process idle timeouts get less racy

`packages/driver/src/child-process.js` already had hard timeouts and idle timeouts. This PR tightens idle timeout behavior by adding a generation guard and a Bun/macOS grace window.

The generation guard makes stale timers harmless:

```js
let idleGeneration = 0;

const resetIdle = () => {
  if (idleTimer) clearTimeout(idleTimer);

  if (effectiveIdleTimeoutMs) {
    idleGeneration += 1;
    const generation = idleGeneration;

    idleTimer = setTimeout(() => {
      if (generation !== idleGeneration) return;
      kill(`CLI idle timed out after ${idleTimeoutMs}ms`, "PROCESS_IDLE_TIMEOUT");
    }, effectiveIdleTimeoutMs);
  }
};
```

That means when stdout/stderr activity resets the watchdog, an older timer cannot wake up later and kill the process anyway.

The tests still preserve the important contract:

```text
hard timeout -> PROCESS_TIMEOUT
idle timeout -> PROCESS_IDLE_TIMEOUT
stdout activity -> resets idle timer
abort signal -> PROCESS_ABORTED
```

## 4. Several tests become more honest instead of just greener

A few test changes are about making tests reflect the real runtime cost and isolation needs:

```diff
- "test": "bun test tests"
+ "test": "bun test --timeout=60000 --max-concurrency=1 tests"
```

That package-script change appears in places where the tests are integration-heavy and share process/runtime state. The point is not to hide failures; it is to stop asking long-running workflow tests to pass under Bun’s short default timeout or unsafe file concurrency.

Other examples:

- `deferred-contract.test.js` uses a larger timer margin so it tests pause/resume behavior instead of scheduler speed under load.
- `heartbeat-bounds.test.jsx` gets a 60s budget and wraps cleanup in `finally`, so failed tests do not leak DB state.
- `workflow-command.e2e.test.js` gets realistic per-test timeouts for CLI subprocess workflows.
- `packages/smithers/tests/e2e-helpers.js` gives CLI subprocesses a default 60s timeout and maps SIGTERM to a stable exit code.

In short, the tests now better match the kind of work they execute.

## 5. Agent diagnostics now wait for the actual report

`packages/agents/tests/agent-diagnostics.test.js` previously checked that `launchDiagnostics("claude", ...)` returned a promise. That left the important work mostly unasserted.

The test now awaits the promise and checks the contents:

```diff
- test("returns a promise for known command", () => {
+ test("returns a diagnostics report for known command", async () => {
    const result = launchDiagnostics("claude", {}, "/tmp");
    expect(result).toBeInstanceOf(Promise);
+   const report = await result;
+   expect(report.agentId).toBe("claude-code");
+   expect(report.checks.length).toBeGreaterThan(0);
  });
```

That is a small change, but it makes the test actually prove diagnostics execute.

## 6. CLI and docs cleanup remove ticket-shaped noise

Some CLI changes are readability cleanup, not behavior changes. The devtools command code keeps the same ideas: preserve command-specific exit codes, route `--json` to the command-level flag, validate devtools argv, and keep telemetry best-effort.

But the comments now explain the current code instead of referencing old “finding #n” ticket language.

Example:

```diff
- // Finding #3: route `--json` to command-scoped `-j` for devtools commands.
  argv = rewriteDevtoolsJsonFlagArgv(argv);
```

The init CTA is also simplified by removing a duplicated `c.agent ? ... : ...` branch where both sides were identical.

Docs also get the private demo packages listed in the package configuration reference:

```diff
+ | `@smithers-orchestrator/smithers-demo` | Private React Flow demo app ... |
+ | `@smithers-orchestrator/smithers-tui-demo` | Private React TUI demo app ... |
```

## The overall story

This PR takes several places that looked finished and makes them more real:

- an exported SQL streaming method now streams
- fork creation is no longer partially durable on failure
- child process watchdogs are less prone to stale timer kills
- async diagnostics tests await real work
- long-running workflow tests get timeouts and isolation that match reality
- CLI comments and docs describe the current system rather than old tickets

So the theme is not “new surface area.” The theme is “make existing surface area trustworthy.”

PR #164, opened 2026-05-26
Add Run Chronicle to the Gateway console

cookesan:
## Summary
- add a Run Chronicle detail view to the built-in Gateway console
- stream the live DevTools tree and run event history over Gateway WebSocket RPCs
- load selected task output and diff data through WebSocket RPC and document the expanded console surface

## Validation
- bun test packages/server/tests/gateway-ui.test.jsx
- pnpm exec oxlint --react-plugin --node-plugin --import-plugin packages/server/src/gatewayUi/defaultOperatorUi.js packages/server/tests/gateway-ui.test.jsx
- git diff --check
- pnpm test

### DIFF
--- a/docs/integrations/gateway.mdx
+++ b/docs/integrations/gateway.mdx
@@ -54,7 +54,7 @@ ws.onopen = () => ws.send(JSON.stringify({
 
 ## Operator Console
 
-Gateway serves a built-in operator console at `/console`. It gives non-coding operators a same-origin surface for health, workflow launch, active runs, and pending approvals. The console uses the Gateway RPC API and accepts the same bearer tokens as programmatic clients.
+Gateway serves a built-in operator console at `/console`. It gives non-coding operators a same-origin surface for health, workflow launch, active runs, pending approvals, and run inspection. Selecting a run opens a Run Chronicle view backed by `streamDevTools`, `streamRunEvents`, `getNodeOutput`, and `getNodeDiff`, so operators can inspect the live task tree, event history, output, and diff without leaving the Gateway. The console uses the Gateway RPC API and accepts the same bearer tokens as programmatic clients.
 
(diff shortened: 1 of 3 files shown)


roninjin10:
Codex review note: this is Codex reviewing, not fucory. fucory will be reviewing immediately after this.

Findings: I did not find a blocking issue in this pass. The PR checks shown on GitHub are green for typecheck, faults, and test. The main residual risk is that most coverage is still asset/string-level for a fairly rich browser UI; a follow-up browser smoke test would make future regressions easier to catch.

Storyline view:

This PR takes the Gateway operator console from a command center into more of a live run journal. Before this, the console could tell an operator what workflows existed, launch a run, list current runs, and handle approvals. That was useful for operating the system, but the moment someone wanted to understand what was happening inside a run, they had to leave the console and use lower-level tooling.

The new flow starts in the same place: the console still loads health, workflows, runs, and approvals through the Gateway RPC surface, and it still keeps bearer auth in session storage rather than local storage. The run list is now interactive, though. Selecting a run opens a Run Chronicle detail area next to the list instead of leaving the run as a static row.

Once a run is selected, the console opens live WebSocket RPC connections. One stream follows DevTools snapshots and deltas, which become the task tree. The other stream follows persisted run events, which become the chronicle. That gives the operator two synchronized views of the same run: the structural tree of tasks and the time-ordered event trail.

The inspector is the last piece. Clicking a tree node, or clicking an event that maps back to a task node, focuses the inspector and loads that task's output and diff through `getNodeOutput` and `getNodeDiff`. So the path through the UI becomes: choose a run, watch its tree and event history, click the interesting task, then inspect the data behind it without leaving `/console`.

The docs update reflects that expanded console surface, and the server UI test now verifies that the built-in asset still parses and includes the new chronicle, WebSocket, stream, output, and diff hooks.

PR #207, opened 2026-06-03
Sync bun lockfile

roninjin10:
Syncs bun.lock with the committed 0.22.0 workspace package versions and the CLI yaml dependency. This keeps Bun's workspace metadata consistent with the package manifests.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -76,7 +76,7 @@
     },
     "apps/cli": {
       "name": "@smithers-orchestrator/cli",
-      "version": "0.21.0",
+      "version": "0.22.0",
       "dependencies": {
         "@clack/prompts": "^0.10.1",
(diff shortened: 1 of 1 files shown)


roninjin10:
Superseded by #214. This PR's only change is a `bun.lock` version resync (0.21.0 → 0.22.0). As part of landing #214 (Postgres/PGlite dialect), `bun.lock` was regenerated authoritatively against the current package manifests — it now carries the 0.22.0 versions **and** the new `pg`/`@electric-sql/pglite` deps. `bun install --frozen-lockfile` on `main` reports no changes, so the lockfile is fully in sync.

This branch predates the pg/pglite deps, so merging it now would strip them back out of `bun.lock` and re-desync it. Closing as done — the sync this PR aimed for is already on `main`.

PR #230, opened 2026-06-09
🐛 fix(pi-plugin): guard inspector render theme

### LINKED ISSUE #225: @smithers-orchestrator/pi-plugin inspector crashes on /smithers in Pi 0.78.1: render() expects theme arg but Pi TUI calls render(width)
## Environment

- Pi: 0.78.1
- `@smithers-orchestrator/pi-plugin`: 0.22.0

## Repro

1. Run `pi`
2. Run `/smithers`
3. Enter a run id, e.g. `run-1780950078277`

## Actual

PI exits due to uncaught exception:

```txt
TypeError: Cannot read properties of undefined (reading 'fg')
    at paint (.../@smithers-orchestrator/pi-plugin/src/views/Header.ts:10:16)
    at Header.render (.../Header.ts:125:9)
    at RunInspector.render (.../RunInspector.ts:124:31)
    at Container.render (.../@earendil-works/pi-tui/dist/tui.js:85:38)
```

## Likely root cause

Smithers inspector components define render methods that expect `theme` as a render argument:

- `Header.render(width, theme)`
- `RunInspector.render(width, height = 34, theme)`

But Pi TUI's component API calls components as:

```ts
render(width: number): string[]
```

So `theme` is undefined at runtime, and `Header.ts` crashes on `theme.fg`.

The extension factory receives `theme` via:

```ts
ctx.ui.custom((_tui, theme, _kb, done) => ...)
```

but the returned `RunInspector` does not appear to close over/store it for later render calls.

## Expected

Opening `/smithers` should not crash PI. The inspector should either close over the theme supplied by `ctx.ui.custom(...)`, or defensively handle missing theme.

jaxernst:
## Summary
- capture the Pi `ctx.ui.custom(...)` theme when constructing the Smithers run inspector
- default inspector renders to that captured theme, or a no-op theme when Pi calls `render(width)` without a theme argument
- make view paint helpers tolerate an absent theme so `/smithers` does not crash on render

Fixes #225.

## Validation
- bun test packages/pi-plugin/tests
- bun run --cwd packages/pi-plugin typecheck
- bunx oxlint --react-plugin --node-plugin --import-plugin -A react/no-children-prop -A unicorn/no-thenable packages/pi-plugin/src/extension.ts packages/pi-plugin/src/views packages/pi-plugin/tests/DevToolsStore.test.ts

### DIFF
--- a/packages/pi-plugin/src/extension.ts
+++ b/packages/pi-plugin/src/extension.ts
@@ -293,6 +293,7 @@ async function openInspector(ctx: ExtensionContext, run: TrackedRun) {
   await ctx.ui.custom((_tui: unknown, theme: unknown, _kb: unknown, done: () => void) =>
     new RunInspector(run.store, run.client, {
       workflowName: run.workflowName,
+      theme: theme as { fg?: (color: string, value: string) => string; bold?: (value: string) => string },
       onClose: done,
       onNotify: (message, level) => ctx.ui.notify(message, level),
     }),
--- a/packages/pi-plugin/src/views/FrameScrubber.ts
+++ b/packages/pi-plugin/src/views/FrameScrubber.ts
@@ -6,8 +6,8 @@ type Theme = {
   bold?: (value: string) => string;
 };
 
-function paint(theme: Theme, color: string, value: string) {
-  return theme.fg ? theme.fg(color, value) : value;
+function paint(theme: Theme | undefined, color: string, value: string) {
+  return theme?.fg ? theme.fg(color, value) : value;
 }
 
 export class FrameScrubber {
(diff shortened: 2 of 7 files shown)


jaxernst:
Closing as superseded by #233 (merged today), which fixes the same crash and closes the underlying issue #225. #233 lands the identical theme-capture mechanism in `RunInspector` (`theme?: Theme` option, `this.theme = options.theme ?? {}`, and `render(width, height, theme = this.theme)`). The only delta here was extra `paint(theme: Theme | undefined, …)` guards in the other view helpers, which aren't needed since those already guard with `theme.fg ? …` and always receive a real theme or `{}`. Thanks!

PR #251, opened 2026-06-11
✨ feat(review): per-file diff narratives + GitHub PR review posting

roninjin10:
Two upgrades to smithers review, demonstrated on this very PR (the tool will review and comment on it):

1. **Per-file diff narratives.** The narrator now writes 2–4 sentences for every file, shown immediately before its diff in the walkthrough — what the diff actually does, which identifiers changed, what to scrutinize — written as a continuous thread within each chapter.

2. **GitHub PR review posting.** `--pr <number|url>` reviews a PR and posts the result onto it: narrative summary as the review body, anchorable findings as inline comments (with suggestion fences), 422-safe fallback, walkthrough link when `--publish` ran.

Spec: `.smithers/specs/smithers-review-walkthrough.md` (Story schema + "GitHub PR reviews" sections).


roninjin10 (commented):
## smithers review — smithers review now writes a per-file narrative paragraph before every diff in the walkthrough and, with --pr, posts the whole review onto a GitHub pull request through the gh CLI.

Two features share this change. First, the story schema grows a per-file narrative field, threaded through the narrate prompt, normalization, the fallback story, and the HTML renderer, so the reader gets two to four sentences of context before each diff instead of a bare role line. Second, an all-new src/github/ pipeline resolves a PR via gh, composes one review (narrative body plus inline comments for every anchorable finding, with a fold-into-body retry), and posts it; the CLI wires --pr to default the diff range to the PR, and the workflow persists the normalized story as a JSON column so the CLI can build the review body after the run. Five of the seven review findings cluster in the new GitHub pipeline and the other two sit in the CLI wiring, so budget review attention accordingly.

**📖 Full walkthrough:** https://review.jjhub.tech/w/xsmjn3pg4ohn

### Reading order

**1. Per-file narratives in the story**

The walkthrough story previously gave each file one role line; this chapter threads a per-file narrative paragraph through every layer that touches a story: schema, agent prompt, normalization, fallback, and HTML rendering. The thing to verify is that the field defaults to an empty string everywhere a story is built without an agent, gets trimmed exactly once in normalizeStory, and is HTML-escaped at render time. No review findings land in these files; the two tests at the end pin both the normalization and the rendering behavior.
- `apps/review/src/walkthrough/storySchema.ts` — Adds the narrative field to the per-file story schema; everything else in this chapter flows from it.
- `apps/review/src/walkthrough/buildNarratePrompt.ts` — Teaches the narrate agent to write the new field.
- `apps/review/src/walkthrough/normalizeStory.ts` — Carries the narrative through story normalization.
- `apps/review/src/walkthrough/fallbackStory.ts` — Keeps the deterministic fallback story shape-compatible.
- `apps/review/src/walkthrough/renderWalkthroughHtml.ts` — Renders the narrative as prose above each diff.
- `apps/review/tests/normalizeStory.test.ts` — Pins narrative trimming and the empty-string default in normalization.
- `apps/review/tests/renderWalkthroughHtml.test.ts` — Pins narrative rendering and escaping in the HTML output.

**2. The GitHub posting pipeline**

An all-new src/github/ directory: a gh CLI runner, two read helpers that resolve the PR and list its changed paths, a pure composer that turns the story plus findings into one review payload, and the poster with its fold-back retry. Read the files in the order listed; each consumes the one before it. Five of the seven review findings live here: a stdin crash risk in runGh, the github.com-only URL parse in resolvePullRequest, hunk-blind anchoring and a breakable suggestion fence in the composer, and an over-broad retry catch in the poster.
- `apps/review/src/github/runGh.ts` — The execFile wrapper every gh call in this chapter goes through.
- `apps/review/src/github/resolvePullRequest.ts` — Resolves a PR number or URL to owner, repo, refs, and head sha.
- `apps/review/src/github/listPullRequestFiles.ts` — Fetches the PR's changed paths for anchor filtering.
- `apps/review/src/github/buildPullRequestReview.ts` — The pure composer: story plus findings in, one review payload out.
- `apps/review/src/github/postPullRequestReview.ts` — Posts the payload via gh api, with a fold-into-body retry.
- `apps/review/tests/buildPullRequestReview.test.ts` — Unit-tests the composer's split, body sections, and size cap.

**3. CLI and workflow wiring**

This is where --pr becomes behavior: the flag parses, the workflow persists the normalized story as a JSON output column, and main.ts stitches together PR resolution, range defaulting, the run, findings re-hydration, composition, and posting. The remaining two review findings are both in main.ts and both concern the range plumbing: the never-fetched head sha and the interaction of --pr with an explicit --commit or --from/--to. Verify the post-run path reads the persisted story column rather than re-deriving the story.
- `apps/review/src/cli/parseReviewArgs.ts` — Adds the --pr flag to argument parsing.
- `apps/review/src/workflow/createReviewWorkflow.tsx` — Persists the normalized story so the CLI can reuse it post-run.
- `apps/review/src/cli/main.ts` — Orchestrates resolve, range defaulting, run, and post.

**4. Docs catch up**

Documentation for both features, kept per the repo's docs-define-the-contract rule. The spec is the authoritative description and the README is the user-facing summary; both describe the fold-back retry and the head-must-exist-locally requirement that the main.ts finding challenges, so if that finding drives an automatic fetch, both files need a follow-up edit.
- `.smithers/specs/smithers-review-walkthrough.md` — Spec for the per-file narrative field and the --pr posting flow.
- `apps/review/README.md` — User-facing usage and pipeline description updates.

_0 finding(s) · 0 inline · generated by smithers review_

roninjin10 (commented):
anchor diagnostic (will be dismissed)

apps/review/src/cli/main.ts
```diff
@@ -46,6 +74,19 @@ async function main() {
...
   const dbPath = args.db ? resolve(args.db) : join(repoDir, ".smithers-review", "review.db");
+
+  let pr: PullRequestTarget | null = null;
+  if (args.pr) {
+    pr = await resolvePullRequest(repoDir, args.pr);
+    if (!args.from && !args.to && !args.commit) {
+      // Prefer the remote-tracking base: the local base branch may be stale
+      // and would drag unrelated commits into the review diff.
+      args.from = refExists(repoDir, `origin/${pr.baseRefName}`) ? `origin/${pr.baseRefName}` : pr.baseRefName;
+      args.to = pr.headSha;
```
roninjin10, line 90: `pr.headSha` is used as the diff target without ever being fetched or verified locally. `resolvePullRequest` only runs `gh pr view --json ...` (no fetch), and `refExists` is applied to the base ref only. When the PR branch has not been fetched into this clone — the common case when reviewing someone else's PR without checking it out — every downstream `git diff <from>..<headSha>` in the workflow fails with "bad object". Fetch `pull/<n>/head` (works for fork PRs too) when the sha is missing.

Relatedly, diffing from `origin/<base>` two-dot diverges from GitHub's PR diff (which is merge-base three-dot): if the base branch has advanced since the PR branched, the reviewed range includes unrelated upstream changes — exactly what the code comment says it is trying to avoid — and findings can anchor to lines outside the PR diff, which makes GitHub reject the whole posted review (422). Diff from `git merge-base <base> <headSha>` instead.

```suggestion
if (!args.from && !args.to && !args.commit) {
      if (!refExists(repoDir, pr.headSha)) {
        execFileSync("git", ["fetch", "origin", `pull/${pr.number}/head`], { cwd: repoDir, stdio: "pipe" });
      }
      const base = refExists(repoDir, `origin/${pr.baseRefName}`) ? `origin/${pr.baseRefName}` : pr.baseRefName;
      // Diff from the merge base so the range matches GitHub's PR diff even
      // when the base branch has advanced since the PR branched.
      let mergeBase = "";
      try {
        mergeBase = execFileSync("git", ["merge-base", base, pr.headSha], { cwd: repoDir, stdio: "pipe" }).toString().trim();
      } catch {}
      args.from = mergeBase || base;
      args.to = pr.headSha;
    }
```

apps/review/src/cli/main.ts
```diff
@@ -46,6 +74,19 @@ async function main() {
...
+    }
```
roninjin10, line 92: This log line runs even when the user combined `--pr` with `--commit` or explicit `--from`/`--to`, in which case the range-assignment branch above is skipped. With `--pr 123 --commit abc` it prints a misleading `→ ..` (both defaults are empty strings). More substantively, in those combinations the review findings are computed from a diff that is not the PR diff, yet the review is still posted at `pr.headSha` — inline line anchors from the foreign diff are likely rejected by GitHub, failing the whole post. Consider scoping the log to the branch that actually set the range, and warning (or erroring) when `--pr` is combined with an explicit range/commit.

apps/review/src/github/buildPullRequestReview.ts
```diff
@@ -0,0 +1,102 @@
...
+  story: Story;
+  findings: Finding[];
+  prPaths: Set<string>;
+  headSha: string;
+  walkthroughUrl?: string;
+}): PullRequestReviewPayload {
+  const anchored: PullRequestReviewPayload["comments"] = [];
+  const unanchored: Finding[] = [];
+  for (const finding of args.findings) {
+    if (finding.startLine > 0 && args.prPaths.has(finding.path)) {
```
roninjin10, line 53: Anchoring is gated only on `startLine > 0` and path membership in `prPaths`, but GitHub's create-review API requires each inline comment's `line` (and `start_line` for multi-line comments) to fall inside the new-side diff hunks for that file. If any single anchored comment misses a hunk (a finding resolved to an unchanged/context line, a line drifted from fallback matching, or a path whose PR status is `removed` so it has no RIGHT side), GitHub rejects the entire review with HTTP 422 ("line must be part of the diff") — dropping the body and every valid comment, which defeats the documented "nothing is dropped" guarantee. Consider passing anchorable new-side line ranges per path (derivable from the `patch` field already fetched in listPullRequestFiles) instead of a bare `Set<string>`, and demoting out-of-range findings to the unanchored list; alternatively, have postPullRequestReview catch the 422 and retry with all comments folded into the body.

apps/review/src/github/buildPullRequestReview.ts
```diff
@@ -0,0 +1,102 @@
...
+const MAX_BODY_CHARS = 60_000;
+const MAX_FILES_PER_CHAPTER = 30;
+
+function findingBody(finding: Finding): string {
+  const parts = [finding.content];
+  if (finding.suggestionCode) {
+    parts.push("", "```suggestion", finding.suggestionCode, "```");
+  }
+  return parts.join("\n");
+}
```
roninjin10, line 29: The suggestion block uses a fixed three-backtick fence, so any `suggestionCode` that itself contains a triple-backtick run (common when the finding targets a markdown file or code containing fenced examples) terminates the `suggestion` block early. The result is a mangled comment, and the truncated suggestion is one click away from being committed by a reviewer via "Apply suggestion". Use a fence longer than the longest backtick run in the content.

```suggestion
function findingBody(finding: Finding): string {
  const parts = [finding.content];
  if (finding.suggestionCode) {
    const runs = finding.suggestionCode.match(/`+/g) ?? [];
    const fence = "`".repeat(Math.max(3, ...runs.map((r) => r.length + 1)));
    parts.push("", `${fence}suggestion`, finding.suggestionCode, fence);
  }
  return parts.join("\n");
}
```

apps/review/src/github/postPullRequestReview.ts
```diff
@@ -0,0 +1,38 @@
...
+  const post = async (body: PullRequestReviewPayload) => {
+    const raw = await runGh(repoDir, ["api", "--method", "POST", endpoint, "--input", "-"], JSON.stringify(body));
+    return JSON.parse(raw) as { html_url?: string };
+  };
+
+  try {
+    const result = await post(payload);
+    return { url: result.html_url ?? pr.url, inline: payload.comments.length };
+  } catch (error) {
+    if (payload.comments.length === 0) throw error;
```
roninjin10, line 26: The fold-into-body fallback fires on *any* error, but the doc comment only justifies it for GitHub's 422 anchoring rejection. Two concrete failure modes slip through this catch-all:

1. **Duplicate reviews.** `post()` does `JSON.parse(raw)` after `runGh` resolves — if the POST succeeded but the response body isn't parseable (or response handling otherwise throws), the review already exists on the PR, and the fallback posts a second one.
2. **Silent degradation / error masking.** A transient failure (auth, rate limit, network, 5xx) downgrades every inline comment into body text when the retry happens to succeed; when it fails too, the second error replaces the original, more informative one.

`runGh` embeds gh's stderr in the thrown error (`gh api failed: HTTP 422: Unprocessable Entity …`), so you can gate the fallback on the 422 signature and rethrow everything else.

```suggestion
} catch (error) {
    const message = error instanceof Error ? error.message : String(error);
    const anchorRejected = /HTTP 422|Unprocessable Entity/i.test(message);
    if (payload.comments.length === 0 || !anchorRejected) throw error;
```

roninjin10 (commented):
## smithers review — smithers review now writes a per-file narrative paragraph before every diff in the walkthrough and, with --pr, posts the whole review onto a GitHub pull request through the gh CLI.

Two features share this change. First, the story schema grows a per-file narrative field, threaded through the narrate prompt, normalization, the fallback story, and the HTML renderer, so the reader gets two to four sentences of context before each diff instead of a bare role line. Second, an all-new src/github/ pipeline resolves a PR via gh, composes one review (narrative body plus inline comments for every anchorable finding, with a fold-into-body retry), and posts it; the CLI wires --pr to default the diff range to the PR, and the workflow persists the normalized story as a JSON column so the CLI can build the review body after the run. Five of the seven review findings cluster in the new GitHub pipeline and the other two sit in the CLI wiring, so budget review attention accordingly.

**📖 Full walkthrough:** https://review.jjhub.tech/w/xsmjn3pg4ohn

### Reading order

**1. Per-file narratives in the story**

The walkthrough story previously gave each file one role line; this chapter threads a per-file narrative paragraph through every layer that touches a story: schema, agent prompt, normalization, fallback, and HTML rendering. The thing to verify is that the field defaults to an empty string everywhere a story is built without an agent, gets trimmed exactly once in normalizeStory, and is HTML-escaped at render time. No review findings land in these files; the two tests at the end pin both the normalization and the rendering behavior.
- `apps/review/src/walkthrough/storySchema.ts` — Adds the narrative field to the per-file story schema; everything else in this chapter flows from it.
- `apps/review/src/walkthrough/buildNarratePrompt.ts` — Teaches the narrate agent to write the new field.
- `apps/review/src/walkthrough/normalizeStory.ts` — Carries the narrative through story normalization.
- `apps/review/src/walkthrough/fallbackStory.ts` — Keeps the deterministic fallback story shape-compatible.
- `apps/review/src/walkthrough/renderWalkthroughHtml.ts` — Renders the narrative as prose above each diff.
- `apps/review/tests/normalizeStory.test.ts` — Pins narrative trimming and the empty-string default in normalization.
- `apps/review/tests/renderWalkthroughHtml.test.ts` — Pins narrative rendering and escaping in the HTML output.

**2. The GitHub posting pipeline**

An all-new src/github/ directory: a gh CLI runner, two read helpers that resolve the PR and list its changed paths, a pure composer that turns the story plus findings into one review payload, and the poster with its fold-back retry. Read the files in the order listed; each consumes the one before it. Five of the seven review findings live here: a stdin crash risk in runGh, the github.com-only URL parse in resolvePullRequest, hunk-blind anchoring and a breakable suggestion fence in the composer, and an over-broad retry catch in the poster.
- `apps/review/src/github/runGh.ts` — The execFile wrapper every gh call in this chapter goes through.
- `apps/review/src/github/resolvePullRequest.ts` — Resolves a PR number or URL to owner, repo, refs, and head sha.
- `apps/review/src/github/listPullRequestFiles.ts` — Fetches the PR's changed paths for anchor filtering.
- `apps/review/src/github/buildPullRequestReview.ts` — The pure composer: story plus findings in, one review payload out.
- `apps/review/src/github/postPullRequestReview.ts` — Posts the payload via gh api, with a fold-into-body retry.
- `apps/review/tests/buildPullRequestReview.test.ts` — Unit-tests the composer's split, body sections, and size cap.

**3. CLI and workflow wiring**

This is where --pr becomes behavior: the flag parses, the workflow persists the normalized story as a JSON output column, and main.ts stitches together PR resolution, range defaulting, the run, findings re-hydration, composition, and posting. The remaining two review findings are both in main.ts and both concern the range plumbing: the never-fetched head sha and the interaction of --pr with an explicit --commit or --from/--to. Verify the post-run path reads the persisted story column rather than re-deriving the story.
- `apps/review/src/cli/parseReviewArgs.ts` — Adds the --pr flag to argument parsing.
- `apps/review/src/workflow/createReviewWorkflow.tsx` — Persists the normalized story so the CLI can reuse it post-run.
- `apps/review/src/cli/main.ts` — Orchestrates resolve, range defaulting, run, and post.

**4. Docs catch up**

Documentation for both features, kept per the repo's docs-define-the-contract rule. The spec is the authoritative description and the README is the user-facing summary; both describe the fold-back retry and the head-must-exist-locally requirement that the main.ts finding challenges, so if that finding drives an automatic fetch, both files need a follow-up edit.
- `.smithers/specs/smithers-review-walkthrough.md` — Spec for the per-file narrative field and the --pr posting flow.
- `apps/review/README.md` — User-facing usage and pipeline description updates.

_7 finding(s) · 7 inline · generated by smithers review_

apps/review/src/github/resolvePullRequest.ts
```diff
@@ -0,0 +1,40 @@
...
+  ]);
+  const data = JSON.parse(raw) as {
+    number: number;
+    url: string;
+    baseRefName: string;
+    headRefName: string;
+    headRefOid: string;
+  };
+  const match = /github\.com\/([^/]+)\/([^/]+)\/pull\//.exec(data.url);
+  if (!match) throw new Error(`cannot parse owner/repo from PR url: ${data.url}`);
```
roninjin10, line 30: The owner/repo regex is hardcoded to `github.com`, so this throws for GitHub Enterprise hosts even though the rest of the flow (the `gh` CLI, including `GH_HOST`) supports them. Since `gh pr view --json url` always returns a canonical `https://<host>/<owner>/<repo>/pull/<n>` URL, matching any host keeps the parse just as safe while removing the github.com-only restriction. The error path is loud, so this is a functional limitation rather than silent corruption, but it blocks `--pr` entirely for GHE users.

```suggestion
const match = /^https?:\/\/[^/]+\/([^/]+)\/([^/]+)\/pull\//.exec(data.url);
  if (!match) throw new Error(`cannot parse owner/repo from PR url: ${data.url}`);
```

apps/review/src/github/runGh.ts
```diff
@@ -0,0 +1,23 @@
...
+          reject(new Error(`gh ${args.slice(0, 2).join(" ")} failed: ${stderr.trim() || error.message}`));
+          return;
+        }
+        resolvePromise(stdout);
+      },
+    );
+    if (stdin != null) {
+      child.stdin?.write(stdin);
+      child.stdin?.end();
+    }
```
roninjin10, line 24: Writing to `child.stdin` without an `error` listener can crash the whole process. If `gh` is not installed (spawn fails with ENOENT) or exits before consuming stdin (bad args, early failure), the pending `write()` fails with EPIPE/ERR_STREAM_DESTROYED, which is emitted as an `error` event on the stdin stream. The `execFile` callback's `reject` does not catch stream errors — an unlistened `error` event throws an uncaught exception and takes down the CLI instead of surfacing the rejected promise. Since this helper feeds the PR review body via stdin (a payload large enough to hit a closed pipe), attach a no-op error handler on stdin and let the execFile callback report the real failure. `end(stdin)` also covers write+end in one call.

```suggestion
if (stdin != null) {
      child.stdin?.on("error", () => {});
      child.stdin?.end(stdin);
    }
```

PR #291, opened 2026-06-16
🐛 fix(agents): run pi json mode non-interactively + capture pi usage (#284)

### LINKED ISSUE #284: PiAgent JSON mode should be non-interactive and complete on terminal pi events
## Summary

`@smithers-orchestrator/agents` `PiAgent` switches to pi JSON mode whenever Smithers passes `onEvent`, but the generated argv omits pi's non-interactive `--print` flag in JSON/RPC modes.

Pi documents these as separate concerns:

```text
--mode <mode>    Output mode: text (default), json, or rpc
--print, -p      Non-interactive mode: process prompt and exit
```

So `--mode json` selects NDJSON output but does not itself make pi process one prompt and exit. A Smithers Pi task can emit a complete `agent_end` / valid final JSON and then remain open as an interactive session until Smithers kills it at the task timeout. Since the base CLI agent harvests the final result on process exit, the completed output can be discarded as `PROCESS_TIMEOUT`.

## Version / environment

- `smithers-orchestrator`: observed during 0.23.0 dogfooding; source paths below are unchanged on current `smithersai/smithers@main` as of `6c2c2a97`
- pi: 0.79.1
- Bun: 1.2.9 / 1.3.x
- OS: macOS (Darwin arm64)

## Evidence in current source

`packages/agents/src/PiAgent.js`:

```js
if (params.mode === "text") {
  if (this.opts.print !== false)
    args.push("--print");
}
else {
  args.push("--mode", params.mode);
}
```

`resolveMode()` returns JSON when Smithers passes `onEvent`:

```js
if (options?.onEvent)
  return "json";
```

So the normal Smithers event-capture path builds `pi --mode json ...`, not `pi --print --mode json ...`.

`BaseCliAgent.generate()` waits for `runCommandEffect()` to complete before flushing interpreter output and returning the result. `PiAgent.createOutputInterpreter()` only emits the final completed event from `onExit`, not from `agent_end`.

## Repro signal

Local pi help confirms `--mode` and `--print` are independent:

```sh
pi --help
# --mode <mode>    Output mode: text (default), json, or rpc
# --print, -p      Non-interactive mode: process prompt and exit
```

A downstream dogfood run saw a Pi task emit complete schema-valid JSON in the transcript, then emit an OSC terminal-title escape and sit idle until Smithers killed it with `PROCESS_TIMEOUT`. The output was not persisted because the process never exited cleanly.

## Expected

- PiAgent JSON/RPC task executions should be non-interactive by default, just like text mode.
- If pi emits a terminal event such as `agent_end`, Smithers should be able to harvest/complete from that event instead of requiring process exit.
- If project extensions can keep a `--print --mode json` process alive after `agent_end`, PiAgent should either disable extensions for non-interactive task execution by default or force-close after the terminal event.

## Actual

- Text mode adds `--print`.
- JSON/RPC mode adds only `--mode <mode>`.
- The completed answer is only surfaced from `onExit`, so an otherwise completed JSON-mode pi run can become a timeout failure.

## Related adapter cleanup

Two lower-priority observations from the same dogfood run likely belong with this adapter fix rather than as separate issues:

1. Pi diagnostics are launched as `launchDiagnostics("pi", env, cwd)`. In the observed failure for a pi invocation configured with `--provider openai-codex`, diagnostics checked Google/Gemini environment (`GOOGLE_API_KEY`, `GEMINI_API_KEY`, `gcloud`) instead of the effective pi provider/model.
2. Pi usage/event telemetry should be verified end-to-end. The run detail showed zero token usage and `eventCount: 0` even though pi JSON output included model events. The current code has generic stdout usage extraction and `onEvent` plumbing, but the Pi interpreter does not attach usage to its completed event.

## Suggested direction

- Build JSON mode argv as `--print --mode json` by default when `this.opts.print !== false`.
- Consider the same for RPC mode if the pi RPC transport expects a single non-interactive task lifecycle.
- Teach `PiAgent.createOutputInterpreter()` to recognize `agent_end` as terminal and capture answer/usage there, not only in `onExit`.
- Make diagnostics provider-aware for pi by inspecting the effective `--provider`/`--model` argv or PiAgent options.

roninjin10:
Fixes #284.

## Problem

`PiAgent` switches to pi's JSON output mode whenever Smithers passes `onEvent`, but `buildArgs` added pi's `--print` flag **only in text mode**. pi documents `--mode` and `--print` as independent concerns:

```
--mode <mode>    Output mode: text (default), json, or rpc
--print, -p      Non-interactive mode: process prompt and exit
```

So `--mode json` selects NDJSON output but does **not** make pi process one prompt and exit. A json-mode Pi task could emit a complete `agent_end`, then sit as an open interactive session until Smithers killed it at the task timeout, and the finished output got discarded as `PROCESS_TIMEOUT`. Separately, completed pi runs reported **zero token usage**.

## What I verified against real pi (0.60.0)

- `--print` and `--mode` are independent flags (confirmed in `pi --help`).
- A json-mode run **with `--print`** exits cleanly (exit 0); the same path drives `PiAgent.generate({ onEvent })` to completion in ~2.5s instead of hanging.
- pi reports final usage under `message.usage` on the assistant's `message_end` with its own field names: `{input, output, cacheRead, cacheWrite, totalTokens}` — not Anthropic's `input_tokens`. The shared `extractUsageFromOutput` only understood Anthropic/Codex shapes, so it returned the all-zero `message_start` and reported 0 tokens.
- pi emits **one `agent_end` per auto-retry attempt** (saw 4 on a rate-limited run) — see "Deliberately not done" below.

## Fixes

1. **`--print` for every non-interactive mode** (`PiAgent.buildArgs`). Text and json/stream-json now both run non-interactively; RPC keeps its own persistent transport (`--mode rpc`, no `--print`). `print: false` still opts out.
2. **pi usage extraction** (`extractUsageFromOutput`) — additive branch that sums usage across assistant `message_end` events; intentionally skips `turn_end` (mirrors the last message → would double count). Claude/Codex extraction unchanged.
3. **Docs** — the pi integration doc literally described the buggy behavior ("JSON/RPC modes set `--mode` and omit `--print`"); corrected + llms bundles regenerated.

End-to-end after the fix: `generate({ onEvent })` → text `"hi"`, events fire, `usage = {inputTokens: 1228, outputTokens: 2, totalTokens: 1230}`, no timeout.

## Deliberately not done (with reasoning)

- **"Treat `agent_end` as terminal / force-close after it."** pi emits one `agent_end` per retry attempt, so completing on the first would truncate a run mid-retry and capture an error result. With `--print`, **process exit** is the correct terminal signal and the existing `onExit` harvesting works. Forcing early completion would regress retried runs.
- **Provider-aware pi diagnostics** (the issue's lower-priority cleanup #1). The pi strategy hardcodes Google auth/rate-limit checks regardless of `--provider`. Fixing it well needs provider plumbing through `launchDiagnostics` + a provider→check mapping across all pi-supported providers — a separate, lower-confidence change best done on its own. Left out to keep this PR focused and low-risk. Happy to follow up.

## Verification

`pnpm -C packages/agents typecheck` ✓ · full agents suite **236 pass / 0 fail** (incl. new pi `--print` matrix + pi usage cases) ✓ · `check:docs` ✓ · llms bundles regenerated.

### DIFF
--- a/apps/cli/docs/llms-full.txt
+++ b/apps/cli/docs/llms-full.txt
@@ -11092,7 +11092,7 @@ const pi = new PiAgent({
 </Task>
 ```
 
-`PiAgent` supports all PI CLI flags: provider/model, tools, extensions, skills, prompt templates, themes, sessions. Text mode uses `--print` by default; JSON/RPC modes set `--mode` and omit `--print`.
+`PiAgent` supports all PI CLI flags: provider/model, tools, extensions, skills, prompt templates, themes, sessions. Text and JSON modes both run non-interactively with `--print` (pi's `--mode` only selects the output format, so JSON also needs `--print` to process one prompt and exit instead of idling until the task timeout); JSON mode adds `--mode json`. RPC mode drives its own persistent stdin/stdout transport, so it sets `--mode rpc` and omits `--print`. Pass `print: false` to opt back into interactive behavior.
 
 PI sessions are first-class hijack targets. `bunx smithers-orchestrator hijack RUN_ID --target pi` reopens the PI session for local steering.
 
--- a/docs/integrations/pi-integration.mdx
+++ b/docs/integrations/pi-integration.mdx
@@ -24,7 +24,7 @@ const pi = new PiAgent({
 </Task>
 ```
 
-`PiAgent` supports all PI CLI flags: provider/model, tools, extensions, skills, prompt templates, themes, sessions. Text mode uses `--print` by default; JSON/RPC modes set `--mode` and omit `--print`.
+`PiAgent` supports all PI CLI flags: provider/model, tools, extensions, skills, prompt templates, themes, sessions. Text and JSON modes both run non-interactively with `--print` (pi's `--mode` only selects the output format, so JSON also needs `--print` to process one prompt and exit instead of idling until the task timeout); JSON mode adds `--mode json`. RPC mode drives its own persistent stdin/stdout transport, so it sets `--mode rpc` and omits `--print`. Pass `print: false` to opt back into interactive behavior.
 
 PI sessions are first-class hijack targets. `bunx smithers-orchestrator hijack RUN_ID --target pi` reopens the PI session for local steering.
 
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -11092,7 +11092,7 @@ const pi = new PiAgent({
 </Task>
 ```
 
-`PiAgent` supports all PI CLI flags: provider/model, tools, extensions, skills, prompt templates, themes, sessions. Text mode uses `--print` by default; JSON/RPC modes set `--mode` and omit `--print`.
+`PiAgent` supports all PI CLI flags: provider/model, tools, extensions, skills, prompt templates, themes, sessions. Text and JSON modes both run non-interactively with `--print` (pi's `--mode` only selects the output format, so JSON also needs `--print` to process one prompt and exit instead of idling until the task timeout); JSON mode adds `--mode json`. RPC mode drives its own persistent stdin/stdout transport, so it sets `--mode rpc` and omits `--print`. Pass `print: false` to opt back into interactive behavior.
 
 PI sessions are first-class hijack targets. `bunx smithers-orchestrator hijack RUN_ID --target pi` reopens the PI session for local steering.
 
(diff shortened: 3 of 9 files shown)


roninjin10:
Superseding this in favor of #292, which is the more complete fix for #284 (adds `agent_end`-terminal handling and provider-aware diagnostics on top of the `--print` + usage changes). Closing as superseded. (Note for #292: worth verifying the `agent_end` harvesting against pi's auto-retry path, since pi emits one `agent_end` per retry attempt.)

PR #444, opened 2026-06-19
fix(cli): avoid hardcoded /bin/bash for agent detection

urchin-tidebot:
## Summary
- replace shelling out through hardcoded `/bin/bash` during agent CLI detection with direct executable lookup on `PATH`
- add a regression test to ensure agent detection does not reintroduce `/bin/bash`

## Why
Hardcoding `/bin/bash` breaks on non-FHS systems like NixOS, where Bash may be available on `PATH` but not at `/bin/bash`. In that case `smithers init` can incorrectly report `binary:codex:no` even when `codex` is installed and authenticated.

## Verification
- `bun test --timeout=120000 --max-concurrency=1 apps/cli/tests/cli-agent-detection.test.js`
- `./node_modules/.bin/tsc -p apps/cli/tsconfig.json --noEmit`
- manual smoke: `bun /home/agent/projects/smithers/apps/cli/src/index.js init --no-skill --force` succeeds on NixOS with Codex installed/authenticated

Requested by @shazow.

### DIFF
--- a/apps/cli/src/agent-detection.js
+++ b/apps/cli/src/agent-detection.js
@@ -1,7 +1,7 @@
-import { existsSync } from "node:fs";
+import { constants, accessSync, existsSync } from "node:fs";
 import { homedir } from "node:os";
 import { join, resolve } from "node:path";
-import { spawnSync } from "node:child_process";
+
 import { SmithersError } from "@smithers-orchestrator/errors";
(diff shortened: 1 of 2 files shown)


shazow:
Nice PR number for baby's first PR. 

image

roninjin10:
Let's go!

roninjin10:
Nix compat

Issue #450, opened 2026-06-23
feedback: installation process with docs


stevennevins:
prompt: 
`Can you find the smithers repo and docs. investigate how to install it and use with my codex subscription`

It seemed to know based on the transcript it should have used my codex subscription in it's plan but the setup resulted in it using the OpenAI api key route for the installation. It was an easy fix after I read the docs and suggested:

```
did you need to run:
bunx smithers-orchestrator agents add --provider codex ... 
```

It might need more language in the docs for using the codex subscription

stevennevins:
Probably a duplicate of https://github.com/smithersai/smithers/issues/448

roninjin10:
Will add an eval for this

PR #451, opened 2026-06-23
🐛 fix(agents): honor Codex CLI subscription auth in Task preflight

### LINKED ISSUE #448: CodexAgent Task preflight requires OPENAI_API_KEY despite configDir subscription auth
CodexAgent subscription auth works with direct codex exec and with CodexAgent.generate({ configDir }), but fails inside workflow <Task>
  preflight.

  Docs say configDir sets CODEX_HOME and no OPENAI_API_KEY is needed. In 0.25.0, <Task> execution calls Codex diagnostics that unconditionally
  require OPENAI_API_KEY.

  Repro:

  const codex = new CodexAgent({
    model: "gpt-5.5",
    configDir: "/Users/me/.codex",
    skipGitRepoCheck: true,
  });

  <Task id="smoke" agent={codex} output={outputs.output}>
    Return JSON.
  </Task>

  Result:

  codex: api_key_valid=fail: OPENAI_API_KEY not set

  Expected: skip API-key diagnostics when configDir/CODEX_HOME subscription auth is configured.

roninjin10:
## Problem

Fixes #448.

`CodexAgent` subscription auth (`configDir` / `CODEX_HOME`, no `OPENAI_API_KEY`) works with direct `codex exec` and with `CodexAgent.generate({ configDir })`, but **fails inside a workflow `<Task>`**. `<Task>` execution runs Codex preflight diagnostics, and the codex `api_key_valid` check unconditionally failed when `OPENAI_API_KEY` was unset:

```
codex: api_key_valid=fail: OPENAI_API_KEY not set
```

The docs already promise that `configDir` sets `CODEX_HOME` and no `OPENAI_API_KEY` is needed — the diagnostic just didn't honor it.

## Fix

Mirror the `codex` binary's own auth resolution in the `api_key_valid` diagnostic. When `OPENAI_API_KEY` is absent, read `<CODEX_HOME>/auth.json` (defaulting to `~/.codex`, honoring `$HOME`):

- ChatGPT subscription tokens (`tokens.access_token`) → **pass** (subscription mode)
- a stored API key (`OPENAI_API_KEY` in the file) → **pass**
- no usable credentials (file missing/unreadable/empty) → **fail**, with an actionable message (`run \`codex login\` or set OPENAI_API_KEY`)

An env `OPENAI_API_KEY` still takes precedence and is probed against the API as before. This also fixes the more general case of a user who ran `codex login` without `configDir` and without an env key.

The check is gated behind a `codexCliAuth` flag so the OpenAI-compatible **pi** strategy (which does not read codex's `auth.json`) keeps requiring the env key.

## Tests

- New `codex strategy checks` cases: subscription tokens pass, stored API key passes, empty `auth.json` fails, no auth fails, env key still probed (precedence).
- Existing "fails when no key" and the `BaseCliAgent preflight` test now isolate `CODEX_HOME` to an empty temp dir so they're deterministic regardless of the developer's real `~/.codex` (previously they would falsely pass on a machine with `codex login`).
- `packages/agents` diagnostics/preflight suites: 39 pass. Root `pnpm typecheck` clean.

### DIFF
--- a/packages/agents/src/diagnostics/getDiagnosticStrategy.js
+++ b/packages/agents/src/diagnostics/getDiagnosticStrategy.js
@@ -1,4 +1,7 @@
 import { spawnSync } from "node:child_process";
+import { readFileSync } from "node:fs";
+import { homedir } from "node:os";
+import { join } from "node:path";
 /** @typedef {import("./DiagnosticCheck.ts").DiagnosticCheck} DiagnosticCheck */
 /** @typedef {import("./DiagnosticCheckId.ts").DiagnosticCheckId} DiagnosticCheckId */
 /** @typedef {import("./DiagnosticContext.ts").DiagnosticContext} DiagnosticContext */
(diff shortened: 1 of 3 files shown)


roninjin10:
### Codex review

Ran `codex review` (gpt-5, reasoning=high) against the base on each iteration:

1. **First pass** — one `[P2]`: a stored API key from `CODEX_HOME/auth.json` was accepted on presence without validation, so an invalid/exhausted stored key could falsely pass preflight (while the spawned `codex` would then fail on the same bad key).
2. **Fix** — refactored to a shared `resolveOpenAiCredentials()`: subscription tokens pass (can't be probed cheaply, matching the Claude subscription check), but a concrete key (env **or** stored in `auth.json`) is probed against `GET /v1/models` exactly like an env key. The rate-limit check probes the same resolved key.
3. **Re-review** — clean: *"No actionable correctness issues were identified in the diff. The changes consistently extend Codex diagnostics to consider Codex CLI auth while preserving pi's environment-key behavior."*

`pnpm typecheck` clean; `packages/agents` diagnostics + pi + preflight suites: 40 pass.

roninjin10:
Fixes https://github.com/smithersai/smithers/issues/450

PR #453, opened 2026-06-23
🔒 fix(server): add Origin allow-list to token/jwt gateway auth

### LINKED ISSUE #446: Gateway: Origin allow-list (`allowedOrigins`) only exists for trusted-proxy mode, not token/jwt
## Summary

The Gateway's Origin allow-list config, `allowedOrigins`, is only available on the **`trusted-proxy`** auth variant. The `token` and `jwt` auth modes — which is what a browser-facing `smithers ui` / operator-UI deployment actually uses — have **no Origin field and no Origin gate** anywhere. Operators running a localhost/LAN gateway in token or jwt mode therefore cannot restrict which web origins may connect, even as defense-in-depth.

## Evidence

- `packages/server/src/GatewayAuthConfig.ts:23` — `allowedOrigins?: string[]` is declared **only** on the `trusted-proxy` variant. The `token` (lines 4–7) and `jwt` (lines 8–19) variants have no such field.
- `packages/server/src/gateway.js` — `authenticateRequest()` only checks `Origin` inside the `trusted-proxy` branch (~lines 3302–3311). Both chokepoints (HTTP RPC via `handleHttpRpc`/`handleElectricWrite`, and the WS upgrade path `listen()` → `handleSocket` → `handleConnect` → `authenticate`) funnel through `authenticateRequest`, so neither gates Origin in token/jwt mode.

Note: the `trusted-proxy` Origin enforcement itself works correctly and is covered by a passing test (`packages/server/tests/gateway.test.jsx`, "trusted-proxy mode enforces allowed origins"). This issue is specifically about the **missing** equivalent for token/jwt.

## Risk

A drive-by web page in a user's browser can probe a localhost/LAN gateway and open a WebSocket / issue RPCs. Bearer tokens in headers mitigate classic CSRF (an attacker can't read the token to set the header), but there is currently no way for an operator to add an Origin allow-list as defense-in-depth for token/jwt deployments. The audit's "Theme B" flagged the WS upgrade as lacking an Origin gate.

## Proposed fix (safe, opt-in, backward-compatible)

- Add optional `allowedOrigins?: string[]` to the `token` and `jwt` variants of `GatewayAuthConfig`.
- Enforce it uniformly at the top of `authenticateRequest()` (before the mode switch): when the configured mode has a non-empty `allowedOrigins`, reject any request/WS upgrade whose `Origin` header is present and not in the list; allow requests with no `Origin` (server-to-server/CLI); when unset/empty, behavior is unchanged (allow all).
- This is opt-in, so it cannot break existing deployments or the smithers UI (operators that opt in include their own UI origin), and it dedupes with the existing `trusted-proxy` check.
- Add a test pinning: configured allow-list accepts matching / rejects non-matching Origin (HTTP + WS); unset allows any; missing Origin allowed.

## Provenance

Found during the test-coverage blind-spot audit (`audit/TEST-COVERAGE-AUDIT.md`, Theme B). The audit's original wording ("`allowedOrigins` enforced nowhere") was a false positive — it *is* enforced for trusted-proxy; this issue is the accurate, narrower residual gap.

roninjin10:
## Problem

Fixes #446.

The Gateway's Origin allow-list (`allowedOrigins`) only existed on the **`trusted-proxy`** auth variant. The `token` and `jwt` modes — what a browser-facing `smithers ui` / operator-UI deployment actually uses — had **no Origin field and no Origin gate**. An operator running a localhost/LAN gateway in token or jwt mode could not restrict which web origins may connect, even as defense-in-depth. A drive-by web page in a user's browser could probe a localhost/LAN gateway and open a WebSocket / issue RPCs.

## Fix (opt-in, backward-compatible)

- Add optional `allowedOrigins?: string[]` to the `token` and `jwt` variants of `GatewayAuthConfig` (`trusted-proxy` already had it).
- Enforce it **uniformly** at the top of `authenticateRequest()`, before the mode switch. Both chokepoints (HTTP RPC via `handleHttpRpc`/`handleElectricWrite`, and the WS upgrade path → `handleConnect` → `authenticate`) funnel through `authenticateRequest`, so both are now gated. When the configured mode has a non-empty `allowedOrigins`, a request/WS upgrade whose browser `Origin` header is present and not on the list is rejected; requests with **no** `Origin` (server-to-server / CLI) are allowed. Unset/empty is unchanged (allow all).
- Dedupe the now-redundant `trusted-proxy` Origin check. This also relaxes trusted-proxy to allow no-Origin requests, matching the uniform CSRF model (the allow-list guards browser drive-by, not non-browser callers); the existing trusted-proxy test still passes.

Opt-in, so it cannot break existing deployments or the smithers UI — operators that opt in include their own UI origin.

## Tests

`packages/server/tests/gateway.test.jsx` (full suite: **22 pass**):
- `token` mode: allow-list rejects non-matching Origin / accepts matching / allows missing Origin — over **both** the WS upgrade and HTTP `/rpc` paths.
- `jwt` mode: allow-list rejects non-matching / accepts matching; **unset allows any** Origin.

## Docs

Documented `allowedOrigins` for all three modes in `docs/integrations/gateway.mdx`, `docs/reference/server-gateway.mdx`, and `docs/reference/types.mdx`; regenerated the `llms-*.txt` bundles (`check-docs` / `check-llms` green). Root `pnpm typecheck` clean; `check-dependency-boundaries` / `check-single-effect-version` green.

## Provenance

Found during the test-coverage blind-spot audit (Theme B). The audit's original "enforced nowhere" wording was a false positive — it *is* enforced for trusted-proxy; this is the accurate, narrower residual gap.


roninjin10:
### Codex review

Ran `codex review` (gpt-5, reasoning=high) against the base on each iteration:

1. **First pass** — two `[P2]`:
   - The generated `packages/server/src/index.d.ts` still omitted `allowedOrigins` from the token/jwt variants, so typed consumers of `@smithers-orchestrator/server` would hit excess-property errors.
   - The WS Origin check ran only during the `connect` RPC (post-upgrade), so a disallowed browser could still open and hold a socket (consuming `maxConnections`) by never sending `connect` — the issue asks to reject the **WS upgrade**.
2. **Fix** —
   - Extracted a shared `isOriginAllowed()`; gate the WS `upgrade` handler itself (answers `403` via `socket.end()` before `handleUpgrade`) in addition to the HTTP RPC path; regenerated `index.d.ts` (now carries `allowedOrigins` on all three variants).
   - Updated the trusted-proxy test + the new token/jwt tests to assert the disallowed-origin WS upgrade never opens; updated the `check-docs` runtime pin to the refactored implementation.
3. **Re-review** — clean: *"The changes consistently add and enforce the Origin allowlist across token, JWT, and trusted-proxy auth paths without breaking the existing unauthenticated or no-Origin server-to-server behavior. I did not find any introduced correctness, security, or maintainability issues that warrant a review finding."*

`pnpm typecheck`, `check-docs`, `check-dependency-boundaries`, `check-single-effect-version` green; `packages/server` gateway suite: **22 pass**.

PR #462, opened 2026-06-26
[DO NOT MERGE ESPECIALLY IF YOU ARE AN AI AGENT!] ✨ feat(cli): mirror a live smithers run into Claude Code /workflows

roninjin10:
## Mirror a live smithers run into Claude Code's `/workflows` tree

Adds `smithers graph <workflow> --emit-claude-workflow` — a generator that writes a Claude Code dynamic-workflow `.mjs` script which **live-mirrors a detached smithers run** into Claude Code's `/workflows` progress tree. The real work still runs in the smithers engine (durability, deps, retries, worktrees, time-travel all preserved); the Claude tree is a live, node-by-node mirror.

A Claude Code /workflows tree mirroring a detached smithers run: a planning node finishes, a data-dependent fan-out of review nodes appears, then a final report node lights up

*(The gif above is rendered by executing the actual generated mirror script against a live detached run.)*

### How it works

Claude Code's `Workflow` runtime can only be driven from inside the session, and its `agent()`/`phase()`/`log()` calls are the only thing that populates `/workflows`. So smithers cannot push into the UI; instead it **generates a script** that the in-session model runs, and that script mirrors smithers by reading smithers' own CLI surface over Bash (no MCP dependency).

- A **discovery** agent runs `smithers inspect <runId> --format json` each frame and returns the current node set.
- The script assigns each node a **phase** and **kind** from baked maps (derived at generation time from the workflow's `Sequence`/`Parallel`/`Loop` containers), mapping runtime loop/fan-out ids (`logicalId@@ralphId=iter`) back to their logical id with a pure `nodeId.split("@@")[0]` lookup.
- One **watcher** agent per node polls `smithers node <id> --run-id <runId> --format json` until terminal, then returns the node's output (or `[skipped]` / `[failed: ...]`).
- Discovery is gated on the next `FrameCommitted` (via `smithers events --type frame --json --watch`) with a monotonic frame cursor, so the loop advances and never busy-spins.

### What ships

- `packages/graph`: a pure `deriveClaudeWorkflowPhases(snapshot)` function (+ types) that turns a `GraphSnapshot` into an ordered phase list and a per-node `{ nodeId, label, phase, kind }` mapping, walking the real `xml` container tree. Unit-tested over real `extractGraph` fixtures (sequence/parallel/ralph, duplicate titles, `@@` ids, every kind).
- `apps/cli`: the `--emit-claude-workflow [--out] [--mirror-all-nodes] [--collapse-phases]` flags, a deterministic template emitter, a path resolver, and a standalone `mirrorState.js` module (parse `inspect`, terminal detection, node-set diff, `FrameCommitted` detection) that is the tested oracle for the agent contract.
- `docs/examples/claude-workflow-mirror.mdx` + the demo gif. Regenerated llms bundles are idempotent.

### Design constraints honored

- **Append-only**: a skipped/vanished node resolves its watcher as a `[skipped]` row (documented).
- **Caps**: a 950-watcher budget and a 5000-frame backstop, both with `log()` on hit (no silent truncation). `--collapse-phases` emits one watcher per phase for large runs.
- **Dynamic by default**: nodes discovered at runtime that the static frame-0 graph could not predict (data-dependent fan-out) are mirrored by default; known compute/static nodes need `--mirror-all-nodes`.
- **Continue-as-new** is not followed; the mirror logs and stops at the original run id.
- **Deterministic, sandbox-safe** output: no `Date.now`/`Math.random`, no absolute paths, no fs/network/Node in the generated body; `meta` is a pure literal and is never referenced from the body.

### Testing

- Unit: phase derivation; mirror-state parse/terminal/diff/frame-signal incl. the dynamic-default and `@@`-suffix paths.
- Generator e2e: deterministic byte-identical output, valid JS, no absolute paths, string-args normalization, no `meta` reference, final-frame mirroring, and collapse-mode membership re-derivation.
- Real-run e2e: seeds a detached run with a **data-dependent `<Parallel>` fan-out** and a fake agent, then drives `mirrorState.js` with real `inspect`/`events` output, asserting it discovers the multi-node fan-out across frames and detects terminal state.
- **Interactive verification**: the generated script was run through the real Claude Code `Workflow` runtime against live detached runs. It mirrored all nodes including the dynamic fan-out and the final report node. This surfaced four runtime-only defects now fixed and covered by tests (args delivered as a JSON string, `meta` not exposed to the body, dynamic nodes dropped by the default kind filter, final-frame nodes dropped on terminal).

### Review

Built with `/codex` planning → independent plan review → implementation → dual review (codex + self) → interactive testing. Codex's review caught a frame-cursor busy-spin blocker and a stale `index.d.ts`; interactive testing caught the four runtime defects above and a collapse-phases late-node gap. All resolved with tests.

Gate: `pnpm typecheck`, `check-docs`/`check-llms`/`check-dependency-boundaries`/`check-single-effect-version`, `packages/graph` (157) and the `claude-workflow` cli suites (13) all green.


roninjin10:

PR #468, opened 2026-06-29
Add versioned llms docs artifacts

SamuelLHuber:
## Summary
- package `docs/llms.txt` and `docs/llms-full.txt` into `smithers-orchestrator`
- add version-pinned docs artifacts (`llms-v<version>.txt` and `llms-full-v<version>.txt`) plus release checks for them
- make `smithers docs` and `smithers docs-full` fetch the installed package version by default, with `--docs-version latest|<semver>|<tag>` overrides

## Validation
- `bun test apps/cli/tests/docs-url.test.js apps/cli/tests/docs-cli-overview-coverage.test.js apps/cli/tests/docs-public-surface-coverage.test.js`
  - `docs-url.test.js`: pass
  - `docs-public-surface-coverage.test.js`: pass
  - `docs-cli-overview-coverage.test.js`: fails in this environment because the spawned CLI cannot resolve workspace imports (`Cannot find module '@smithers-orchestrator/engine' from apps/cli/src/index.js`)
- `cd packages/smithers && npm pack --dry-run`
  - confirms the tarball now includes `docs/llms.txt` and `docs/llms-full.txt`

### DIFF
--- a/apps/cli/docs/llms-full.txt
+++ b/apps/cli/docs/llms-full.txt
@@ -51,12 +51,12 @@ Smithers orchestrates AI coding agents at scale with composable, model- and harn
 <Note>
 This page is in the **For Agents** track. It is for AI harnesses and workflow authors who
 need the runtime model. Human-facing docs live in the **For Humans** Guide, starting at
-[What Smithers Is](/guide/what-is-smithers).
+What Smithers Is.
 </Note>
 
 Smithers runs your workflow on a render loop: each frame it asks "what has finished, and what can start now?" Tasks produce outputs validated by Zod schemas; the runtime persists them to SQLite. Crashes, restarts, and approvals are first-class, and the runtime resumes from the last persisted state without re-running completed work.
 
-**Default authoring path - MDX prompts.** For most workflows you edit a plain Markdown file in `.smithers/prompts/` and the seeded TypeScript skeleton picks it up automatically. No build step, no TypeScript required. See [MDX Workflow Authoring](/guides/mdx-authoring).
+**Default authoring path - MDX prompts.** For most workflows you edit a plain Markdown file in `.smithers/prompts/` and the seeded TypeScript skeleton picks it up automatically. No build step, no TypeScript required. See MDX Workflow Authoring.
 
 **Advanced authoring - TypeScript SDK.** When you need branching, schemas, parallel fan-out, or loops, write the workflow as a JSX tree. Smithers renders the tree and drives execution:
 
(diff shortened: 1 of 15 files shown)


SamuelLHuber:
initial prompt / idea. aka the why from me.

goal being that the cli and agents always get the correct version of the docs / agent skill that they have installed locally so one isn't forced to always be on the latest version

----

"is llms-full.txt versioned so we can get the one for the exact smithers version? so we could bundle it?"

```
Yes in git, no in the published package.

`docs/llms-full.txt` is a committed artifact generated from the MDX docs by [`scripts/generate-llms.ts`](/Users/samuel/git/playground/smithers/scripts/generate-llms.ts:1), and it exists in the `v0.21.0` tag, so you can get the exact file for an exact Smithers source version from the repo/tag history. The generator explicitly writes [`docs/llms-full.txt`](/Users/samuel/git/playground/smithers/scripts/generate-llms.ts:272).

But `smithers-orchestrator` does not currently ship `docs/` in npm. Its package manifest only includes `src/` in `files` at [`packages/smithers/package.json`](/Users/samuel/git/playground/smithers/packages/smithers/package.json:49). So `node_modules/smithers-orchestrator/docs/llms-full.txt` is not something you can rely on today, even though the PI plugin tries that path as a fallback at [`packages/pi-plugin/src/extension.ts`](/Users/samuel/git/playground/smithers/packages/pi-plugin/src/extension.ts:78).

Also, the CLI `smithers docs-full` command fetches the current published site artifact from `https://smithers.sh/llms-full.txt`, not a version-pinned one, as shown in [`apps/cli/src/index.js`](/Users/samuel/git/playground/smithers/apps/cli/src/index.js:5215).

So the practical answer is:

- Exact-by-tag/commit from git: `yes`
- Exact-by-installed-npm-version today: `no`
- Bundleable with the package today: `not unless we change packaging`

If you want, I can make the packaging change so each published `smithers-orchestrator` version includes `docs/llms-full.txt`.
```

"does it show how the docs are served so could we do a surgical pr with minimal changes that enables this?"

"okay so first we package the llms-full.txt into the npm okay, but then do we have agent skills that are versioned? and on the docs i want to have a llms-full-<tag>.txt too so one where we serve the latest and one where we can go by semver tag. how would we accomplish that"

"on the cli it shouhld also open the version that it is, otehrwise there maybe version mismatches. 

now wire all that re versioned docs, cli and npm packaging into a feature branch and open a draft pr using our fork origin (see fork)."

PR #471, opened 2026-07-01
Fix child subflow output selection

### LINKED ISSUE #456: Subflow childRun returns undefined when child output schema is not named output
I ran into a `Subflow mode="childRun"` output issue where the child run finishes successfully and persists its output rows, but the parent receives `undefined` and then fails output validation.

Minimal repro: https://github.com/ralfboltshauser/smithers-subflow-output-repro

In the repro:

- child writes to `childOutput`
- parent runs it through `<Subflow mode="childRun" output={outputs.childOutput} />`
- child run finishes
- child output rows are persisted
- parent fails with `Task output failed validation for childOutput` because the subflow result is `undefined`

From a quick source read, it looks like `Subflow` returns `RunResult.output`, and `RunResult.output` is only populated from a child schema table named exactly `output`.

Expected: child workflows should be able to declare/select their workflow-level output explicitly, instead of requiring a magic schema key named `output`.

One possible minimal fix:

```ts
const childWorkflow = child.smithers(fn, {
  output: child.outputs.childOutput,
});
```

Then Subflow can use that declared child workflow output and store it into the parent output target.

cookesan:
## Summary
- add a workflow-level output option that controls final RunResult output selection
- resolve child subflow results from the declared output target instead of requiring a table named `output`
- keep existing behavior for workflows that still use an output table named `output`
- cover a child workflow writing `childOutput` into a parent `result` output

Fixes #456.

## Tests
- `bun test packages/engine/tests/workflow-make-contract.test.jsx`
- `pnpm --filter @smithers-orchestrator/engine typecheck`
- `pnpm --filter @smithers-orchestrator/scheduler typecheck`
- `pnpm --filter @smithers-orchestrator/engine test`
- `pnpm -r typecheck`
- `pnpm check:deps`
- `pnpm check:effect`

### DIFF
--- a/apps/cli/docs/llms-full.txt
+++ b/apps/cli/docs/llms-full.txt
@@ -4369,18 +4369,29 @@ const child = createSmithers({
   childResult: z.object({ summary: z.string() }),
 });
 
-const childWorkflow = child.smithers((ctx) => (
-  <child.Workflow name="scan-repo">
-    <Sequence>
-      <Task id="scan" output={child.outputs.childResult} agent={scanner}>
-        {`Scan ${ctx.input.repo} and summarize.`}
-      </Task>
-    </Sequence>
-  </child.Workflow>
-));
+const childWorkflow = child.smithers(
+  (ctx) => (
+    <child.Workflow name="scan-repo">
+      <Sequence>
+        <Task id="scan" output={child.outputs.childResult} agent={scanner}>
+          {`Scan ${ctx.input.repo} and summarize.`}
+        </Task>
+      </Sequence>
+    </child.Workflow>
+  ),
+  { output: child.outputs.childResult },
+);
 
 export default childWorkflow;
 ```
+
+The second argument to `smithers(build, { output })` declares which schema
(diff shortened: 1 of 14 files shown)


roninjin10:

**Independent Opus review (LGTM):** Fixes a real, still-present gap on main — `RunResult.output` (and every Subflow childRun result) was only ever read from a child schema key literally named `output`, breaking the pattern our own subflow.mdx documents. The opt-in `smithers(build, { output })` resolution mirrors task-level output resolution at the single site where `RunResult.output` is computed, covers both child-run paths, and keeps default behavior byte-for-byte unchanged. Maintainer improvements applied: uses drizzle's `isTable` to fail fast on unregistered targets (instead of a downstream TypeError), copies `ambiguousZodSchemas` through `resolveChildWorkflow` so the shared-raw-Zod ambiguity error also fires for childRun, the option is typed within the scheduler dependency boundary, and docs/subflow.mdx is fixed with bundles regenerated. New tests use the real engine + SQLite (no mocks) and genuinely fail without the fix. Scope tight, no strays.

**Codex second opinion (verdict: merge).**

PR CI green.

Issue #300, opened 2026-06-17
🏛️ [audit] CI enforcement, architecture & systemic policy (34)


roninjin10:
Repo-wide policy/enforcement issues: the high bar is not enforced by CI, plus the systemic root-cause defect classes (checkJs off, committed .d.ts drift, duplicated helpers, the observability publish cycle) and dependency-boundary gaps. Fixing these once removes whole classes of per-package findings.

Source: [`audit/REVIEW.md`](https://github.com/smithersai/smithers/blob/main/audit/REVIEW.md) (2026-06-16 multi-agent audit, adversarially verified). Full per-finding evidence + recommended fix is in `audit/reconciled-findings.json`.

**34 findings.**

#### `apps/cli` (1)
- [ ] **P2** index.js is a 6,439-line monolith mixing parsing, ~60 command bodies, MCP wiring, and helpers — `apps/cli/src/index.js (6439 lines)`

#### `architecture-deps` (10)
- [x] **P1** dependency-boundary check scans ZERO files for the e2e workspace (directWorkspaceDirs entry is effectively dead) — `scripts/check-dependency-boundaries.mjs:101 (filesForPackage)`
- [ ] **P2** Circular dependency between @smithers-orchestrator/agents and @smithers-orchestrator/observability ships to npm — `packages/agents/src/BaseCliAgent/BaseCliAgent.js:5-6 and apps/observability/src/_traceEventNormalizers.js:1-2`
- [ ] **P2** observability is a foundational library but lives in apps/ (package masquerading as an app) — `apps/observability/package.json`
- [ ] **P2** .smithers workflow pack (a shipping target) is excluded from the dependency-boundary check and imports undeclared react/effect — `scripts/check-dependency-boundaries.mjs:13 (workspaceRoots/directWorkspaceDirs) and .smithers/package.json`
- [ ] **P2** Subpath exports point their `types` condition at the barrel index.d.ts, which does not contain the subpath's symbols — `packages/agents/package.json (exports "./BaseCliAgent".types -> ./src/index.d.ts)`
- [x] **P2** Three workspace packages (accounts, usage, tool-context) are missing from the root tsconfig paths map — `tsconfig.json:24-234 (paths)`
- [ ] **P2** smithers <-> cli package cycle exists (bin delegates dynamically; cli imports smithers statically) — `packages/smithers/src/bin/smithers.js:138 and apps/cli/src/*.js`
[...]

PR #473, opened 2026-07-01
fix(vcs): bound jj repo probe

cookesan:
## Summary
- bound the JJ repo probe with the same 1.5s timeout used by pointer and snapshot calls
- return false on probe timeout so durability startup can continue without a stuck JJ subprocess
- add a fake-JJ regression test for a sleeping repo probe

Part of #300.

## Verification
- `pnpm --filter @smithers-orchestrator/vcs test`
- `pnpm --filter @smithers-orchestrator/vcs typecheck`
- `pnpm -r typecheck`
- `git diff --check origin/main..HEAD`

### DIFF
--- a/packages/vcs/src/jj.js
+++ b/packages/vcs/src/jj.js
@@ -20,9 +20,7 @@ import { Duration, Effect, Fiber, Metric, Stream } from "effect";
 import { vcsDuration } from "@smithers-orchestrator/observability/metrics";
 import { resolveJjBinary } from "./resolveJjBinary.js";
 
-const JJ_POINTER_TIMEOUT_MS = 1_500;
-const JJ_REPO_TIMEOUT_MS = 1_500;
-const WORKSPACE_SNAPSHOT_TIMEOUT_MS = 1_500;
+const JJ_PROBE_TIMEOUT_MS = 1_500;
 /**
  * @param {Stream.Stream<Uint8Array, unknown, never>} stream
  * @returns {Effect.Effect<string, unknown, never>}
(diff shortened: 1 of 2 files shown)


roninjin10:

**Verdict: approve — rebased, improved, merging on green CI.**

**Fable review:** The premise verifies on main: `isJjRepo` pipes `runJj` straight into `Effect.map` with no timeout, while `getJjPointer` and both snapshot calls are bounded at 1.5s — the exact unchecked audit finding in #300. The hot path is real: `startDurability` awaits `isJjRepoFn(cwd)` before every durability start, so a hung jj blocks the agent attempt indefinitely. The regression test is load-bearing (fake jj sleeps 2s > 1.5s bound; without the fix `expect(false)` fails assertively, not via runner timeout) and CI-safe via the established fake-jj shim (Windows-compatible, no real jj). No mocks — real subprocess fault injection.

Maintainer improvements applied on the branch (rebased onto main): consolidated the three identical 1.5s `timeoutTo` sentinel blocks (pointer, snapshot log/op, and the new repo probe) into one `withJjTimeout` helper, and added an `Effect.logWarning` on timeout — previously a timed-out probe silently downgraded durability to the no-op handle with no diagnosable signal. Fable re-review verified the consolidation is behavior-preserving: sentinel stderr strings byte-identical (nothing in the repo asserts them), error channel stays `never`, interrupt kills the child (no process leak), and the `Effect.ensuring` metric still fires on interrupt. 79/79 packages/vcs tests + package/root typecheck green. Verdict: LGTM.

One follow-up worth a future look (out of scope here): the probe command `jj log -r @ -n 1 --no-graph` lacks `--ignore-working-copy`, so on a large dirty worktree the probe itself can trigger a working-copy snapshot — the likeliest way to actually hit this timeout.

**Codex second opinion (verdict: merge):** independently confirmed the unbounded probe on main, that the test "would fail on current main because the sleep exits 0 after 2 seconds and unbounded isJjRepo() would return true", and judged the behavioral tradeoff (slow probe → durability off for that attempt) consistent with the existing 1.5s snapshot/pointer policy. Flagged the 500ms margin between the 2s fake sleep and 1.5s bound as minor residual test risk, matching the existing timeout tests' style.

PR #474, opened 2026-07-01
fix(devtools): include approval and human task kinds

cookesan:
## Summary
- extend the devtools task kind type to include approval and human nodes
- keep the tracked declaration file aligned with the source type

## Why
The protocol package already exposes approval and human task kinds. Pi plugin imports protocol snapshots and uses devtools applyDelta, so the narrower devtools declaration rejects current snapshots during recursive typecheck.

## Tests
- `pnpm --filter @smithers-orchestrator/devtools typecheck`
- `pnpm --filter @smithers-orchestrator/pi-plugin typecheck`
- `pnpm --filter @smithers-orchestrator/devtools test`
- `pnpm -r typecheck`
- `pnpm check:deps`
- `pnpm check:effect`
- `git diff --check`

### DIFF
--- a/apps/cli/docs/llms-full.txt
+++ b/apps/cli/docs/llms-full.txt
@@ -10291,7 +10291,7 @@ type DevToolsNode = {
   type: "workflow" | "task" | "sequence" | "parallel" | /* …see protocol */;
   name: string;
   props: Record<string, unknown>;
-  task?: { nodeId: string; kind: "agent" | "compute" | "static"; /* … */ };
+  task?: { nodeId: string; kind: "agent" | "compute" | "static" | "human" | "approval"; /* … */ };
   children: DevToolsNode[];
   depth: number;
 };
--- a/docs/integrations/gateway.mdx
+++ b/docs/integrations/gateway.mdx
@@ -344,7 +344,7 @@ type DevToolsNode = {
   type: "workflow" | "task" | "sequence" | "parallel" | /* …see protocol */;
   name: string;
   props: Record<string, unknown>;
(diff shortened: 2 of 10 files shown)


roninjin10:

**Independent Opus review (LGTM):** On current main, devtools' `DevToolsNode.task.kind` is still the narrow `agent|compute|static` union while protocol already declares the 5-kind union and the gateway snapshot builder emits `human`/`approval` at runtime — so this widening is the correct root-cause fix. Dropping the pi-plugin bridging assertion is a clean improvement: it typechecks now that the unions align and makes the alignment compiler-enforced (the real regression guard against future drift). Scope is tight — the two type files, the pi-plugin refactor, the gateway.mdx source edit, and the six regenerated bundles it feeds are all required and complete (llms-integrations.txt correctly unchanged, versioned bundle matches v0.26.1), so check-docs/check-llms stay green and no consumer switches exhaustively on the widened union.

Maintainer changes applied on the branch: removed the `Parameters<typeof applyDelta>` bridging casts in pi-plugin's DevToolsStore so `applyDelta(base, delta)` is called directly (a prior consumer sweep confirmed the two live readers — gateway-client's `nodeKind()` and pi-plugin's NodeInspector — already tolerate the new values, and no exhaustive switch exists); fixed the stale documented union in docs/integrations/gateway.mdx and regenerated bundles.

**Codex second opinion (verdict: improve-then-merge):** agreed the fix is correct, and its "improve" ask was exactly the two things done here — enforce the alignment at compile time and update the stale docs.

PR CI green.

PR #475, opened 2026-07-01
test(time-travel): stabilize revert DB tests

cookesan:
## Summary
- give DB-backed revert tests per-test 30s budgets for slower Windows runners
- close each test database from finally so assertion failures do not leak SQLite handles into later tests
- rebase on current main after the global time-travel test timeout landed

## Verification
- `bun test packages/time-travel/tests/revert.test.js`
- `pnpm --filter @smithers-orchestrator/time-travel typecheck`
- `pnpm --filter @smithers-orchestrator/time-travel test`
- `pnpm -r typecheck`
- `git diff --check origin/main..HEAD`

### DIFF
--- a/packages/time-travel/tests/revert.test.js
+++ b/packages/time-travel/tests/revert.test.js
@@ -15,6 +15,9 @@ function buildDb() {
     ensureSmithersTables(result.db);
     return result;
 }
+
+const DB_TEST_TIMEOUT_MS = 30_000;
+
 describe("revertToJjPointer", () => {
     test("calls jj restore with correct pointer", async () => {
         const result = await Effect.runPromise(revertToJjPointer("abc123").pipe(Effect.provide(BunContext.layer)));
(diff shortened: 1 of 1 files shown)


roninjin10:

**Verdict: closing — already implemented on main.**

**Fable review:** The substantive claim — DB-backed revert tests need 30s budgets for slower Windows runners — landed on main in f38603ff7, which added `setDefaultTimeout(30_000)` at the top of the same file. In bun:test that file-level default applies to every test in the file, so this PR's per-test `DB_TEST_TIMEOUT_MS = 30_000` is byte-for-byte redundant (same 30s value, no behavioral change). These tests are green on main's Windows shards today without this PR.

The only residual delta is wrapping each test body in `try/finally { cleanup() }`. The stated justification (leaking SQLite handles into later tests) doesn't hold: `createTestDb` builds a fresh SQLite DB in a fresh `mkdtemp` dir per test and no later test touches that file. The rest is ~140 lines of reindentation churn.

**Codex second opinion (verdict: merge):** Codex agreed the timeout piece is redundant but valued the `try/finally` as real robustness — an assertion failure mid-test would skip `cleanup()` and leave the handle open. That's true as far as it goes, but the leaked handle is on a per-test private temp file in an already-failing run, and the process exit releases it; it doesn't justify the churn. Codex also noted the pre-existing weak success-path assertion in `revert.test.js` (`RevertFinished` is emitted on restore failure too) — that's a fair follow-up for a separate, focused PR.

Thanks for the PR — the timeout race was real when it was authored; main just got there first.

PR #478, opened 2026-07-01
ci: run OpenAPI drift check in CI

cookesan:
## Summary
- add a root `check:openapi` script that delegates to the Gateway package drift guard
- run the OpenAPI drift guard in the Linux typecheck job after Bun is available
- document the new root check and refresh generated llms docs

## Why
The Gateway package already checks that `packages/gateway/openapi.yaml` matches the current RPC registry during package builds. Running the same guard in CI catches stale committed OpenAPI documents earlier, before release or package build paths are the first place to notice drift.

## Verification
- `pnpm check:openapi`
- `pnpm check:llms`
- `node scripts/check-docs.mjs`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/ci.yml")'`
- `pnpm check:deps`
- `pnpm check:effect`
- `pnpm -r typecheck`
- `pnpm --filter @smithers-orchestrator/gateway build`
- `git diff --check`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,11 +26,19 @@ jobs:
           node-version: 22
           cache: pnpm
 
+      - uses: oven-sh/setup-bun@v2.2.0
+        if: runner.os == 'Linux'
+        with:
+          bun-version: 1.3.13
+
       - run: pnpm install --frozen-lockfile
 
       - run: pnpm check:effect
 
       - run: pnpm check:deps
+
+      - run: pnpm check:openapi
+        if: runner.os == 'Linux'
 
       - run: pnpm lint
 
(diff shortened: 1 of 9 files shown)


roninjin10:

**Verdict: closing — protection already implemented on main.**

**Fable review:** OpenAPI drift is already gated per-PR twice on current main: (1) `packages/gateway/tests/generate-openapi.test.ts` (landed in 7735b792b, PR #455) ends with a byte-equality test that regenerates the document via `toYaml(buildOpenApiDocument())` and asserts it matches the committed `openapi.yaml` — logically identical to `generate-openapi.ts --check` — and runs in the required CI test job via `pnpm -r --no-bail test` on Linux, the Windows shards, and coverage. (2) The per-PR **Faults** workflow runs `pnpm -r build`, and the gateway build script is `bun run check:openapi && tsc`, invoking the exact guard this PR wires up (green in this PR's own check rollup).

The mechanical delta is real (main's root package.json lacks `check:openapi`; the typecheck job lacks the step) but the protection is fully redundant, and the cost is adding Bun to the otherwise node-only typecheck job plus 7 llms-bundle files that now conflict with main's refreshed docs.

**Codex second opinion (verdict: improve-then-merge):** Codex reached the same core finding — "OpenAPI drift is already caught on current main by the Gateway test… it is not closing an untested CI gap" — but valued the earlier, named failure signal. It also flagged that the PR would make `docs/examples/ci-workflow.mdx` stale (it documents Bun as test-job-only). We're weighing the faster-signal value below the cost of a redundant gate + doc churn; if a named signal is wanted later, promoting the existing gateway test into a labeled CI step without the root-script/Bun changes would be the cleaner shape.

Thanks — the audit trail (findings 5483/5794) shows this gap was real when identified; it has since been closed at the test and build layers.

PR #479, opened 2026-07-01
fix(cli): keep generated configs and prompt tests stable

cookesan:
## Summary
- add the TypeScript 6 deprecation opt-out to the `.smithers/tsconfig.json` generated by `smithers init`
- make the fuzzy-select test harness keep readline's editable line in sync before emitting keypresses

## Why
Main now has the package-level TypeScript 6 config updates. This keeps initialized workflow packs aligned with that baseline, and makes the prompt regression test exercise the same readline state the interactive prompt reads when filtering and backspacing.

## Verification
- `pnpm -r build`
- `pnpm -r typecheck`
- `node scripts/check-docs.mjs`
- `pnpm --filter @smithers-orchestrator/cli exec bun test tests/init.e2e.test.js`
- `pnpm --filter @smithers-orchestrator/cli exec bun test tests/fuzzy-select.test.js`
- `pnpm --filter @smithers-orchestrator/cli exec bun test --coverage --coverage-reporter=lcov --coverage-dir=/tmp/smithers-cli-coverage --timeout=120000 --max-concurrency=1 tests`
- `pnpm check:deps`
- `pnpm check:effect`
- `git diff --check`

### DIFF
--- a/apps/cli/src/workflow-pack.js
+++ b/apps/cli/src/workflow-pack.js
@@ -187,6 +187,7 @@ function renderTsconfig() {
             strict: true,
             skipLibCheck: true,
             baseUrl: ".",
+            ignoreDeprecations: "6.0",
             paths: {
                 "~/*": ["./*"],
             },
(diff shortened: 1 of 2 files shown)


cookesan:
Closing this because current `main` already removes the generated `baseUrl` entry, so the TS 6 deprecation path this branch patched no longer exists. Keeping the branch now adds only obsolete suppression config. The latest `main` typecheck covers the intended fix.

PR #480, opened 2026-07-01
fix(workflows): type Telegram digest workflow

cookesan:
## Summary
- return the resolved Telegram digest source directly after auto-selection
- reuse a narrowed notable-links payload before mapping digest links
- restore green `.smithers` and workspace typecheck on current main

## Why
The new Telegram digest workflow currently fails `.smithers` typecheck in two places: one dead `"auto"` comparison after source resolution, and one repeated `unknown` expression that TypeScript cannot narrow. This keeps runtime behavior the same while making the authored workflow compile.

## Verification
- `pnpm -r typecheck`
- `pnpm check:deps`
- `pnpm check:effect`
- `pnpm -r build`
- `git diff --check`


cookesan:
Closing since current main already includes the same typecheck fix in this workflow.

PR #481, opened 2026-07-01
fix(cli): handle fuzzy picker backspace

cookesan:
## Summary
- normalize DEL/BS control bytes before fuzzy picker filtering
- keep the readline line buffer aligned after normalization
- make the fake keypress helper include the key metadata that readline normally provides

## Why
The CLI fuzzy picker can keep delete bytes in the tracked query under Bun-backed synthetic TTY input. That leaves the prompt in the no-match validation path after Backspace, so the prompt promise stays pending and the CLI test suite hangs.

## Verification
- `pnpm --filter @smithers-orchestrator/cli exec bun test --timeout=10000 --max-concurrency=1 tests/fuzzy-select.test.js`
- `pnpm --filter @smithers-orchestrator/cli test`
- `pnpm --filter @smithers-orchestrator/cli typecheck`
- `pnpm --filter @smithers-orchestrator/cli build`
- `pnpm -r --no-bail test`
- `git diff --check`

### DIFF
--- a/apps/cli/src/fuzzy-select.js
+++ b/apps/cli/src/fuzzy-select.js
@@ -144,6 +144,26 @@ function terminalRows() {
 }
 
 /**
+ * Some readline implementations keep DEL/BS bytes in `line` for synthetic TTYs.
+ * Treat them as editing commands before filtering so tests and Bun-backed
+ * interactive sessions stay aligned.
+ *
+ * @param {string} input
+ * @returns {string}
+ */
+function normalizeBackspaces(input) {
+    const chars = [];
+    for (const char of String(input ?? "")) {
+        if (char === "\b" || char === "\x7f") {
+            chars.pop();
+        } else {
+            chars.push(char);
+        }
+    }
+    return chars.join("");
+}
+
+/**
  * A type-to-filter picker built on clack's `Prompt` base class. We subclass
  * `Prompt` directly (NOT `SelectPrompt`) with `trackValue=true` so the base
  * pipes readline and keeps `this.value` in sync with the typed query — letters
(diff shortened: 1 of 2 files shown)


roninjin10:

**Verdict: closing — the reported bug is not reproducible on main, and the PR's own tests pass against the unfixed code.**

**Fable review:** The claim is that Bun-backed synthetic TTY input leaves DEL/BS bytes in the fuzzy picker's tracked query, sticking the prompt in the no-match path and hanging the CLI suite. Verified false on current main: (1) `bun test tests/fuzzy-select.test.js` on main (Bun 1.3.14) — 28/28 pass, including 'backspace widens the filter again' and 'Enter on no matches reasks', which would hang exactly as described if DEL bytes stayed in the query; (2) main CI (pinned Bun 1.3.13) shows every fuzzySelect interactive test passing; (3) decisively: the PR's **modified test file run against main's unmodified `src/fuzzy-select.js` passes 28/28** — the tests do not discriminate between fixed and unfixed code. Mechanically, clack core 1.4.2 mirrors `rl.line` into `userInput` only after readline's own `_ttyWrite` processes the key, and both Node's and Bun's `_ttyWrite` handle `key.name === 'backspace'` via delete-left before any self-insert path — a DEL byte cannot reach `rl.line` unless a synthetic event omits the backspace name, which readline's keypress parser never produces. The fix also reaches into clack/readline internals (`this.userInput`, `rl.line`, `rl.cursor`, bypassing `_setUserInput`), which is fragile against clack upgrades.

**Codex second opinion (verdict: improve-then-merge):** Codex independently confirmed "on current main, apps/cli/tests/fuzzy-select.test.js passes" and that it could only reproduce the failure by synthesizing "a DEL keypress without a usable `name`" — i.e. an event readline's parser never emits. It suggested that if this is pursued, it needs a regression test that actually fails on main first. We agree with that bar: without a failing reproduction on a supported input path, the normalization guards an unreachable state at the cost of coupling to library internals.

If you can produce a real reproduction (specific terminal/Bun version emitting nameless DEL keypresses through the production CLI path), please reopen with that failing test — that would change the calculus entirely.

PR #482, opened 2026-07-02
test(e2e): cover public subpath exports

cookesan:
## Summary
- add an e2e runtime guard for each documented `smithers-orchestrator/*` facade subpath
- keep the existing programmatic API barrel coverage in the same export-focused test file
- cover #300 public subpath-resolution gap while keeping e2e typecheck inside package boundaries

Part of #300.

## Verification
- `bun test e2e/exports/programmatic-api.test.ts`
- `pnpm -C e2e typecheck`
- `pnpm -r typecheck`
- `pnpm --filter @smithers-orchestrator/e2e test`
- `pnpm lint`
- `pnpm check:effect`
- `pnpm check:deps`
- `git diff --check`

### DIFF
--- a/e2e/exports/programmatic-api.test.ts
+++ b/e2e/exports/programmatic-api.test.ts
@@ -1,5 +1,10 @@
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
 import { expect, test } from "bun:test";
 import * as smithers from "smithers-orchestrator";
+
+const HERE = dirname(fileURLToPath(import.meta.url));
 
 // Guards the programmatic run-control + output-reading API on the public barrel.
 // These power embedders that drive durable runs across processes (start a run,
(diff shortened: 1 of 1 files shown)


roninjin10:

**Verdict: approve — merging.**

A single e2e test file that guards every documented `smithers-orchestrator/*` facade subpath still resolves at runtime with a non-empty export set — closing the #300 audit item ("No automated guard that every documented public subpath export resolves"; most facades resolve purely by `./*` wildcard file-existence, so a renamed/deleted facade breaks a documented import with no CI signal). I verified the diff directly: it derives the explicit (non-wildcard) subpaths from `packages/smithers/package.json` exports at test time — so newly added facades like `telegram`/`cloudflare` are covered automatically — plus a curated list of the 9 wildcard-backed documented subpaths, and asserts each dynamic import is non-empty. No mocks, CI-safe (real imports), scope is exactly one test file. All three triage improvements applied.

**Codex second opinion (verdict: improve-then-merge; the derive-from-exports improvement was applied).**

PR CI green.

PR #483, opened 2026-07-02
fix(scripts): keep docs normalizer on CLI tokens

cookesan:
## Summary
- tighten the docs CLI normalizer boundary so path, scoped-package, dot-prefixed, and hyphen-prefixed `smithers` tokens stay intact
- export the normalizer helpers for direct regression coverage without scanning docs during import
- run the normalizer regression test from the docs gate

## Verification
- `bun test scripts/normalize-bunx.test.ts`
- `bun scripts/normalize-bunx.ts --check`
- `pnpm check:docs`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm -r typecheck`
- `pnpm check:effect`
- `pnpm check:deps`
- `pnpm check:llms`
- `git diff --check`

### DIFF
--- a/scripts/check-docs.mjs
+++ b/scripts/check-docs.mjs
@@ -177,6 +177,11 @@ const CLOUD_PRODUCT_SPEC = join(root, ".smithers/specs/cloud-execution-product.m
 
 let failed = false;
 
+{
+  const r = spawnSync("bun", ["test", join("scripts", "normalize-bunx.test.ts")], { cwd: root, stdio: "inherit" });
+  if (r.status !== 0) failed = true;
+}
+
 for (const script of ["normalize-bunx.ts", "normalize-placeholders.ts"]) {
   const r = spawnSync("bun", [join("scripts", script), "--check"], { cwd: root, stdio: "inherit" });
   if (r.status !== 0) failed = true;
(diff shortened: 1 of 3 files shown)


roninjin10:

The docs `bunx` normalizer widened its token-boundary lookbehind from `(?<![/\\w])` to `(?<![/\\\\w@.-])` so it stops rewriting `smithers` when it's glued to a non-CLI prefix (Windows paths, scoped names/@mentions, domain-like tokens, hyphenated bins).

**Independent Opus review (LGTM):** Verified the premise by reverting only the lookbehind to main's version — 3 of the PR's 4 test cases then fail (`C:\\tools\\smithers run`, `my-smithers run`, and the hyphenated-bin code span), so the regression test is load-bearing and the fix correct, with no realistic false negatives (markdown bullets and `$ ` prompts put a space before `smithers`). The `import.meta.main` refactor is the correct Bun idiom (importing the module runs no doc-scanning side effects), and wiring `bun test scripts/normalize-bunx.test.ts` into check-docs.mjs is CI-safe because `scripts/` is the workspace root that `pnpm -r test` never covers and the files sit outside the root tsconfig. All three triage improvements applied; scope is exactly the 3 germane files, no strays, no doc content changed (no llms regen needed).

**Codex second opinion (verdict: merge):** independently validated the current-vs-proposed regex behavior from an isolated reproduction and judged risk low (the boundary only reduces false positives).

PR CI green.

PR #484, opened 2026-07-02
test(protocol): guard error code drift

cookesan:
## Summary
- add protocol tests that compare runtime error-code lists with their TypeScript aliases
- compare the generated public declaration tuples with the same runtime lists
- keep the gateway/devtools error contract from drifting silently

## Verification
- `pnpm --filter @smithers-orchestrator/protocol test`
- `pnpm --filter @smithers-orchestrator/protocol typecheck`
- `pnpm --filter @smithers-orchestrator/protocol build`
- `pnpm typecheck`
- `pnpm -r typecheck`
- `pnpm typecheck:examples`
- `pnpm typecheck:evals`
- `pnpm check:deps`
- `pnpm check:effect`
- `pnpm check:docs`
- `pnpm check:llms`
- `pnpm lint` (existing warnings)
- `node scripts/check-smithers-test-script.mjs`
- `git diff --check`

### DIFF
--- a/packages/protocol/tests/protocol-contracts.test.js
+++ b/packages/protocol/tests/protocol-contracts.test.js
@@ -1,7 +1,9 @@
 import { describe, expect, test } from "bun:test";
 import { spawnSync } from "node:child_process";
+import { readFileSync } from "node:fs";
 import { dirname, resolve } from "node:path";
 import { fileURLToPath } from "node:url";
+import ts from "typescript";
 import { DEVTOOLS_PROTOCOL_VERSION } from "../src/devtools.js";
 import {
   DEVTOOLS_ERROR_CODES,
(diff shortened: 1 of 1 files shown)


roninjin10:

This is a single-file, tightly-scoped addition of two drift-guard tests that compare protocol's error-code type aliases and the committed `src/index.d.ts` declaration tuples against the runtime arrays via the TypeScript compiler API. The premise holds on current main (nothing links the three representations, and there's no d.ts-regen CI gate), so the guard is real. No mocks; `typescript` is the package's own devDependency; CI-safe.

**Independent Opus review (LGTM):** Replicated the AST-parsing logic against the real committed sources — all assertions pass. The maintainer-triage improvements were applied correctly: an order-contract comment on ERROR_CODE_CONTRACTS, acceptance of the drift-proof `(typeof CODES)[number]` alias form with a clear failure message (so a legitimate refactor isn't met with a cryptic error), and an early-return in the d.ts visitor. Scope is exactly one test file.

**Codex second opinion (verdict: merge):** independently validated the guard closes a real gap and flagged no blocking issue.

Local + PR CI green (typecheck, all gates, protocol 11/11).

PR #495, opened 2026-07-02
fix(engine): fail timer wake resume mismatches

### LINKED ISSUE #494: Run parked on a Timer dies silently when its workflow source changes: overdue timer shows 'Time remaining: 0s' forever, never surfaces as unhealthy
Observed 2026-07-02 on a looping watchdog run (monitor-smithers, run fleet-medic-0702c):

1. Launch a detached run of a workflow with a `<Ralph>` loop containing a `<Timer duration="5m">`. The run reaches `waiting-timer` and its driver parks.
2. Edit the workflow .tsx (or any file in its import graph) while the run sleeps.
3. The timer comes due, the wake goes through the resume path, resume fails `RESUME_METADATA_MISMATCH` — and nothing surfaces it. `ps` shows `waiting-timer` indefinitely; `why` shows `Fires at: <past> / Time remaining: 0s` 18+ minutes after the fire time. No failed status, no event, no log line an operator would find.

Expected: an overdue timer whose wake cannot proceed should transition the run to failed (with the mismatch error) or at minimum surface as unhealthy/stale in `ps` and `why` (e.g. 'timer overdue by 18m — wake failed: RESUME_METADATA_MISMATCH'). A watchdog that dies this way is indistinguishable from one that is merely sleeping.

Related: the same module-graph guard also bricks explicit `--resume` after editing imports like `.smithers/agents.ts` — that part is by design, but the silent overdue-timer variant means the guard fires with no observer.

cookesan:
## Summary
- fail unattended timer wake resumes when durable metadata no longer matches the parked run
- cancel pending timer state before writing the terminal run failure
- add regression coverage for supervisor and Gateway timer wake paths
- normalize fuzzy picker backspace handling and keep selection state separate from readline cursor state
- keep Gateway and Open Code Review tests synchronized with runtime readiness and token telemetry
- ship the OpenClaw plugin runtime entry that the installer copies into the extension

Fixes #494.

## Tests
- bun test packages/engine/tests/durability.test.jsx
- bun test apps/cli/tests/supervisor-core.test.js
- bun test packages/server/tests/gateway-timers.test.jsx
- bun test packages/engine/tests/deferred-contract.test.js
- bun test apps/cli/tests/fuzzy-select.test.js --timeout=120000 --max-concurrency=1
- bun test apps/cli/tests/agent-wiring.test.js apps/cli/tests/gateway-command.test.js apps/cli/tests/fuzzy-select.test.js apps/cli/tests/published-deps-declared.test.js --timeout=120000 --max-concurrency=1
- bun test ./.smithers/tests/open-code-review.test.ts ./.smithers/tests/open-code-review-ui.e2e.test.ts --timeout=120000 --max-concurrency=1
- pnpm typecheck
- pnpm lint
- pnpm test

### DIFF
--- a/packages/engine/src/engine.js
+++ b/packages/engine/src/engine.js
@@ -1865,6 +1865,55 @@ function assertResumeDurabilityMetadata(existingRun, existingConfig, current, wo
     }
 }
 /**
+ * Claim-owner prefix the supervisor stamps on runs it resumes:
+ * apps/cli/src/supervisor.js builds `supervisor:<supervisorId>` when it claims a
+ * run for an unattended resume. The engine keys "unattended timer wake"
+ * detection off this cross-package contract, so keep the two in sync.
+ */
+const SUPERVISOR_CLAIM_OWNER_PREFIX = "supervisor:";
+/**
+ * An unattended timer wake (the supervisor sweeping a due `waiting-timer` run)
+ * that hits a resume-durability mismatch must fail the run instead of leaving it
+ * parked forever while the sweep retries silently. Interactive `--resume`
+ * mismatches still throw without failing the run.
+ *
+ * @param {RunRow | null | undefined} existingRun
+ * @param {RunOptions} opts
+ * @returns {boolean}
+ */
+function shouldFailTimerWakeResume(existingRun, opts) {
+    if (existingRun?.status !== "waiting-timer")
+        return false;
+    return opts.resumeClaim?.claimOwnerId?.startsWith(SUPERVISOR_CLAIM_OWNER_PREFIX) === true;
+}
+/**
+ * @param {SmithersDb} adapter
+ * @param {EventBus} eventBus
+ * @param {string} runId
+ * @param {unknown} error
+ */
+async function markTimerWakeResumeFailed(adapter, eventBus, runId, error) {
+    const errorInfo = errorToJson(error);
+    await cancelPendingTimers(adapter, runId, eventBus, "run-failed");
+    const failedAtMs = nowMs();
+    await Effect.runPromise(adapter.updateRun(runId, {
+        status: "failed",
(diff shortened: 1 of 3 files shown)


roninjin10:
## Maintainer review (Opus review + Codex second opinion) — landing via batch #510

**Verdict: approved & landed.** Rebased onto main, verified, and merged as part of the linear batch stack #510 (all real CI green: every typecheck + test shard, coverage, faults, postgres).

**Independent Opus review (LGTM):** `shouldFailTimerWakeResume` now fires for BOTH default-active wakers — the supervisor claim path AND the previously-dropped gateway timer-sweep path (`config.gatewayTriggeredBy === "timer:gateway"`), gated on `waiting-timer` status inside the resume-durability-mismatch catch, so it can only fire on a genuine mismatch and never on a clean resume. The bug is real on main: at the mismatch point the run row isn't yet owned by the process, so the engine's outer catch skips DB persistence and gateway's `.catch` only broadcasts a transient failed event — the row stays `waiting-timer` and `processDueTimers` re-sweeps it forever (#494). The new `markTimerWakeResumeFailed` persists `failed` (+ cancelPendingTimers + RunFailed) before rethrowing. Both regression tests are real-backend (one asserts the persisted DB row; one drives the REAL `gateway.processDueTimers` chain, not a monkeypatched resume) and genuinely fail without the fix. Scope tight (engine guard + 2 constants + 2 tests), no strays, no interaction with #471.

**Codex second opinion (verdict: merge).**

Landed on main via the batch. Thanks @cookesan — Co-Authored-By preserved on the commit.

roninjin10:
Landed on main via the linear batch #510 (rebase-merged; your commit is on main with Co-Authored-By preserved). Closing this source PR as merged-equivalent.

PR #497, opened 2026-07-02
fix(cli): ship OpenClaw plugin entrypoint

cookesan:
## Summary
- add the OpenClaw plugin runtime entrypoint declared by package metadata
- register the Smithers tool contracts through the native OpenClaw plugin
- keep the plugin entrypoint visible to Git and assert it is installed
- decouple fuzzy-select option highlighting from Clack's text cursor so backspace filtering stays stable

## Tests
- pnpm --filter ./apps/cli test
- pnpm --filter ./apps/cli typecheck
- pnpm --filter ./apps/cli build
- pnpm typecheck
- pnpm lint

### DIFF
--- a/apps/cli/src/fuzzy-select.js
+++ b/apps/cli/src/fuzzy-select.js
@@ -150,11 +150,17 @@ function terminalRows() {
  * (including j/k/h/l) become filter text, while arrow keys still move the cursor.
  *
  * Instance fields we own (separate from base `this.value`/`this._cursor`):
- *  - `query`      live filter text, mirrored from base `this.value`.
- *  - `allOptions` immutable input options.
- *  - `filtered`   current fuzzyFilter(query, allOptions) result.
- *  - `cursor`     index into `filtered` (clamped), NOT into allOptions.
- *  - `maxItems`   windowing budget.
+ *  - `query`        live filter text, mirrored from base `this.value`.
+ *  - `allOptions`   immutable input options.
+ *  - `filtered`     current fuzzyFilter(query, allOptions) result.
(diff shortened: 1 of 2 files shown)


cookesan:
Follow-up from the fresh rebase: the current red Windows shard is the gateway-react sqlite cleanup `EBUSY` issue, not an OpenClaw entrypoint regression. I opened #500 with the cleanup retry and the gateway autostart scope fix; I’ll rerun/rebase this once that base fix lands or is otherwise available.

roninjin10:

**Verdict: approve — rebased, scope-corrected, merging.**

Note on scope: this PR's original headline (ship the OpenClaw plugin `./dist/index.js` entrypoint) was **subsumed by main commit 388b6b59d**, which landed a tracked source entrypoint (`apps/cli/src/openclaw-plugin/index.js` + toolSpecs) that is functionally superior to the committed dist artifact. That half was stripped on rebase (the dist build-artifact, its .gitignore exception, and the dist test would have landed dead code). What remains is the PR's other, still-unfixed change.

**Independent Opus review (LGTM):** The fuzzy-select fix is real on current main — `apps/cli/src/fuzzy-select.js` uses base-owned `this._cursor` as its option-highlight index, but @clack/core@1.4.2's `Prompt.onKeypress` runs `this._cursor = this.rl.cursor` on every tracked non-return keypress (verified at dist/index.mjs:288), so the highlight gets clobbered — arrow-down sticks near the top and Enter can return a lower-ranked match. The fix migrates all 8 functional usages to an owned `this.optionCursor`; remaining `_cursor` mentions are accurate comment references. Scope is tight: only fuzzy-select.js + its test.

**Codex second opinion (verdict: improve-then-merge).**

PR CI green.

PR #511, opened 2026-07-03
fix(cli): validate preselected restore targets

cookesan:
## Summary
- reject preselected restore checkpoints that do not match the requested node, iteration, or seq
- normalize stray erase characters before filtering fuzzy-select input

## Tests
- bun test apps/cli/tests/snapshots-restore.test.js apps/cli/tests/fuzzy-select.test.js --timeout=120000
- pnpm --filter @smithers-orchestrator/cli typecheck
- pnpm --filter @smithers-orchestrator/cli test

### DIFF
--- a/apps/cli/src/fuzzy-select.js
+++ b/apps/cli/src/fuzzy-select.js
@@ -144,6 +144,23 @@ function terminalRows() {
 }
 
 /**
+ * @param {string} value
+ * @returns {string}
+ */
+function normalizeTrackedInput(value) {
+    let result = "";
+    for (const char of String(value ?? "")) {
+        if (char === "\x7f" || char === "\b") {
+            result = result.slice(0, -1);
+        }
+        else {
+            result += char;
+        }
+    }
+    return result;
+}
+
+/**
  * A type-to-filter picker built on clack's `Prompt` base class. We subclass
  * `Prompt` directly (NOT `SelectPrompt`) with `trackValue=true` so the base
  * pipes readline and keeps `this.value` in sync with the typed query — letters
@@ -199,10 +216,10 @@ class FuzzySelectPrompt extends Prompt {
             if (i >= 0) this.optionCursor = i;
         }
 
-        // Re-filter on every keystroke. Typing AND backspace both edit rl.line,
-        // so the base re-emits "userInput" for both — this covers backspace free.
+        // Re-filter on every keystroke. Normalize stray erase characters because
+        // fake streams and some readline paths can leave them in `rl.line`.
         this.on("userInput", () => {
-            this.query = this.userInput ?? "";
+            this.query = normalizeTrackedInput(this.userInput ?? "");
             this.filtered = fuzzyFilter(this.query, this.allOptions);
             if (this.optionCursor > this.filtered.length - 1) {
                 this.optionCursor = Math.max(0, this.filtered.length - 1);
--- a/apps/cli/src/fuzzy-select.js
+++ b/apps/cli/src/fuzzy-select.js
@@ -144,23 +144,6 @@ function terminalRows() {
 }
 
 /**
- * @param {string} value
- * @returns {string}
- */
-function normalizeTrackedInput(value) {
-    let result = "";
-    for (const char of String(value ?? "")) {
-        if (char === "\x7f" || char === "\b") {
-            result = result.slice(0, -1);
-        }
-        else {
-            result += char;
-        }
-    }
-    return result;
-}
-
-/**
  * A type-to-filter picker built on clack's `Prompt` base class. We subclass
  * `Prompt` directly (NOT `SelectPrompt`) with `trackValue=true` so the base
  * pipes readline and keeps `this.value` in sync with the typed query — letters
@@ -216,10 +199,10 @@ class FuzzySelectPrompt extends Prompt {
             if (i >= 0) this.optionCursor = i;
         }
 
-        // Re-filter on every keystroke. Normalize stray erase characters because
-        // fake streams and some readline paths can leave them in `rl.line`.
+        // Re-filter on every keystroke. Typing AND backspace both edit rl.line,
+        // so the base re-emits "userInput" for both — this covers backspace free.
         this.on("userInput", () => {
-            this.query = normalizeTrackedInput(this.userInput ?? "");
+            this.query = this.userInput ?? "";
             this.filtered = fuzzyFilter(this.query, this.allOptions);
             if (this.optionCursor > this.filtered.length - 1) {
                 this.optionCursor = Math.max(0, this.filtered.length - 1);
--- a/apps/cli/src/restore.js
+++ b/apps/cli/src/restore.js
@@ -90,7 +90,11 @@ export async function runRestoreOnce(opts) {
     if (opts.target) {
         target = pickTargetCheckpoint([opts.target], selection);
         if (!target) {
-            stderr.write(`Preselected checkpoint (node ${opts.target.nodeId} seq ${opts.target.seq}) does not match requested run ${runId} node ${nodeId}${iteration !== undefined ? ` iteration ${iteration}` : ""}${seq !== undefined ? ` seq ${seq}` : ""}\n`);
+            // Include the target's own iteration/seq so an iteration-only
+            // mismatch is legible. Checkpoint rows carry no runId, so the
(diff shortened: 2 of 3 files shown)


roninjin10:
Hi @cookesan — quick status from the PR-cleanup sweep, since this is your active PR.

**This PR (#511) looks good on its own:** tightly scoped (fuzzy-select + restore validation, follow-up to the merged #497/#499), CI green. I'm intentionally **not** merging it as part of this pass — it was opened after the cleanup snapshot, so per the maintainer it's slated for the next-release batch rather than this one.

**Heads-up on main, though:** three commits landed directly on main under `docs:` subjects — `f2a34c95b`, `b2b72b6d3` ("document Cloudflare and Telegram approval, fix gateway/sync drift"), `dd229aff7` — but they actually carry ~198 files / ~9.8k insertions of code (gateway, openapi, scheduler, time-travel, vcs, and a new `packages/gateway-ui`). That push turned main CI red in three deterministic ways:

1. **`packages/gateway-ui/src/NodeRow.tsx:54`** — `.smithers` typecheck fails `TS2322: Property 'key' does not exist on type 'NodeRowProps'`. gateway-ui is authored for React JSX (`jsxImportSource: react`) but exports its types as source `.tsx`, so `.smithers` (which uses the `smithers-orchestrator` JSX runtime) recompiles `NodeRow.tsx` and rejects the React `key` on the recursive `<NodeRow key={…}/>` introduced when NodeRow was extracted from RunTree (`3a2647d79`).
2. **seeded-pack drift** — `apps/cli/tests/seeded-pack-fresh.test.js` fails; `.smithers` was edited without re-running `bun scripts/generate-workflow-pack.ts`.
3. **`apps/cli/tests/legacy-workflow-output.e2e.test.js:89`** — the validation-loop "implement" workflow output now yields `1` where the test expects `0`.

Everything from the actual PR sweep (16 merged, 4 closed) is intact and was green at the stack merge `5e68c285b`; this red is purely from the later push. **I'm landing minimal fixes now to get main green.** For next time, labeling these as `feat/fix` (not `docs`) and running `pnpm -r typecheck` + the seeded-pack regen before pushing would catch all three locally. Thanks!

cookesan:
Thanks for the careful sweep and for keeping the next-release boundary clear. I’ll keep #511 scoped as-is.

roninjin10 (commented):

Good catch on the core issue: `runRestoreOnce` trusted a caller-preselected `target` blindly, so a stale or mis-reported checkpoint would be reverted even when it didn't match the requested node/seq. I rebased onto `main` and tightened the fix:

### Reworked: dedupe the match logic against the listing path
Your version added a second predicate (`matchesCheckpointSelection`) that duplicated the filter inside `pickTargetCheckpoint` — two copies of "does this checkpoint match the selection?" that can silently drift (e.g. if the tie-break or a new selector is added to one but not the other). I dropped the duplicate and validate the preselected row by running it through the existing picker: `pickTargetCheckpoint([opts.target], selection)`. One predicate, one source of truth.

### Fixed: misleading error message on a rejected preselect
Previously a rejected preselected target fell through to `No matching durability checkpoint for run … node …`, which reads like a *listing* miss and hides the real problem (you passed a target that doesn't match what you asked for). It now emits a distinct, actionable message: `Preselected checkpoint (node … seq …) does not match requested run … node … [iteration …] [seq …]`.

### Tests
- Updated the rejected-node test to assert the new distinct message.
- Added a **mismatched-seq rejection** test (right node, wrong `--seq` → rejected, nothing reverted).
- Added a **matching-seq acceptance** test (preselected target that matches `--seq` → honored, reverts the right commit).

### Dropped
- The bundled `apps/cli/src/fuzzy-select.js` backspace-normalization hunk. That "bug" doesn't reproduce under the locked `@clack/core@1.4.2` (readline strips DEL before clack mirrors `rl.line`); it only appears with a stale local `@clack/core` on disk. Reverted to `main`; the fuzzy suite stays green (30/30).

`pnpm typecheck` and `bun test apps/cli/tests/snapshots-restore.test.js` (11/11) pass. Fable subagent re-review next; merging once it's clean.

roninjin10:
**Verdict: mergeable — no blocking or major issues; a few minor nits below.**

Re-review of the reworked branch at head `6350102`.

## BLOCKING / MAJOR

None found.

The core mechanism is sound: `pickTargetCheckpoint([opts.target], selection)` (`apps/cli/src/restore.js:91`) reuses the exact listing predicate, so preselect validation and listing selection cannot diverge by construction. `filter()` allocates a fresh array, so the caller's row is never mutated and the sort is a no-op on ≤1 element; the row is returned iff it survives the same nodeId/iteration/seq filters as the listing path. I traced the `Number()` coercions: a valid preselect cannot be falsely rejected (the only in-tree caller that passes `target` — `restore_checkpoint` in `apps/cli/src/mcp/semantic-tools.js:1456-1477` — picks it with the identical selection immediately beforehand, and re-filtering a picked row with the same selection is idempotent), and an invalid one is only accepted along dimensions the request left unconstrained, which matches the listing path's "latest for the node" semantics.

## MINOR / nits

1. **Rejection message omits the target's iteration** — `apps/cli/src/restore.js:93`. The parenthetical prints only `node ... seq ...`. For an iteration-only mismatch (requested `--iteration 2`, target has `iteration: 1`, same node and seq) the message reads e.g. `Preselected checkpoint (node n1 seq 0) does not match requested run r1 node n1 iteration 2 seq 0` — the printed target looks identical to the request, which is confusing exactly when the user most needs the discriminating field. Suggest including `iteration ${opts.target.iteration}` in the parenthetical.
2. **"does not match requested run ${runId}" slightly overstates the check** — `apps/cli/src/restore.js:93`. `Checkpoint` rows carry no `runId`, so run membership is not (and cannot be) validated here; a cross-run target with coincidentally matching node/seq would pass. Not a regression (the listing path is run-scoped via `listWorkspaceCheckpoints(runId)`, and the only caller passing `target` lists from the same run), but consider wording like `…does not match the requested selection (run ${runId} node …)` or adding `runId` to the row so it can be checked for real.
3. **Test gaps** — `apps/cli/tests/snapshots-restore.test.js`: the three new tests are meaningful and non-tautological (they assert exit code, that `revert` was never invoked, and the distinct message; the acceptance case asserts the actual revert args `["c0", "/wt"]`). Two cheap additions would round them out: (a) an **iteration-mismatch** rejection (all fixture rows have `iteration: 0` and no test passes `iteration` with a preselect), and (b) a **rejected preselect with an adapter also present**, asserting no fallback to listing occurs (`listWorkspaceCheckpoints` never called, exit 1). That second case pins the key contract: fail loudly rather than silently re-pick — which the code gets right (`if/else if`), but nothing currently guards it against a future refactor to a fallback.
4. **Message asymmetry**: the listing-miss message (`apps/cli/src/restore.js:101`) includes `seq` but not `iteration`, while the new preselect message includes both. Pre-existing, but now that `selection` is built once, unifying the two would be trivial.
5. **Stale PR title**: "validate restore targets **and fuzzy input**" — the `fuzzy-select.js` hunk was dropped (confirmed unchanged vs main), so the title's second half no longer describes the diff. Worth trimming before merge.

## Verified

- Net diff vs `origin/main` at `6350102` touches only `apps/cli/src/restore.js` (+17/−4) and `apps/cli/tests/snapshots-restore.test.js` (+66); `apps/cli/src/fuzzy-select.js` is byte-identical to main.
- `bun test tests/snapshots-restore.test.js` at the head SHA (isolated worktree): **11 pass, 0 fail**.
- `matchesCheckpointSelection` does not exist at head or on main — the net diff never introduces the duplicate predicate, so there is no parity question left vs main: the picker itself is unchanged.
- CLI caller (`apps/cli/src/index.js` `restore` command) never passes `target` and parses `--iteration`/`--seq` as zod numbers, so the `Number()` coercions are exercised with real numbers there; the MCP semantic tool is the sole preselect caller and uses a matching selection.
- Rejection path returns `{ exitCode: 1 }` before any `revert`, so no filesystem mutation on mismatch.

---
*Automated fable re-review of head `63501025098f82da49566a6355e70fff6baad18d`.*

roninjin10:
Applied the fable re-review's nits: the rejection message now names the target's own iteration/seq and no longer implies run-membership is validated (checkpoint rows carry no `runId`); added an iteration-mismatch test and a test asserting a rejected preselect never falls back to listing from the adapter. Retitled to drop the stale "fuzzy input" mention. Merging on green CI.

PR #512, opened 2026-07-03
fix(gateway-ui): pin React JSX runtime for consumer workspaces

cookesan:
## Summary

Pin the React JSX runtime for the gateway-ui components. Adds `/** @jsxImportSource react */` to the 11 components that lacked it (NodeRow.tsx already had it), so all 12 declare React explicitly — otherwise a consumer workspace whose default `jsxImportSource` points at `smithers-orchestrator` (as ours does for authoring workflow UIs) mis-compiles these React components.

Scoped to the pragmas: the codex auth-gating and fuzzy-select changes originally bundled here were dropped/deferred (the codex-auth change carries a module-load hang risk and a repo-wide fixture blast radius that belong in a dedicated PR).

## Tests
- `pnpm typecheck`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #513, opened 2026-07-03
fix(packages): publish subpath declarations

### LINKED ISSUE #504: 🐛 graph/integrations: "./*" subpath export maps every subpath type to index.d.ts (untyped for consumers)
## Summary
The `"./*"` subpath export maps every subpath's `types` to the core `index.d.ts`, so consumers importing a subpath get the package's core types, not the subpath's — the subpaths are effectively untyped.

## Location
`packages/graph/package.json` `exports["./*"].types = "./src/index.d.ts"`. The same pattern is still present in `packages/integrations/package.json`.

## Impact
`import { X } from "@smithers-orchestrator/graph/<subpath>"` resolves to the wrong (core index) types for external consumers.

## Suggested fix
Ship real per-subpath types: generate a `.d.ts` next to each importable `.js` via a tsup multi-entry build and map `"./*".types` to `"./src/*.d.ts"`. Watch the same-basename `X.js`/`X.ts` twin-collision documented for the integrations package (the type-twin can win and drop value exports) — rename twins to `<name>Types.ts` if needed.

## Why deferred
This needs a build-config change (per-file `.d.ts` generation), not a one-line `package.json` edit, and the twin-collision risk warrants care.

_From the full-codebase review._

cookesan:
## Summary
- generate per-subpath declaration files for graph and integrations packages
- map wildcard export types to `./src/*.d.ts` so deep imports resolve their own published contracts
- add resolver tests covering both valid imports and intentional type misuse
- set the React JSX runtime for gateway UI TSX files so `.smithers` typecheck handles React `key` props correctly
- refresh the generated seeded workflow pack used by CLI coverage
- keep CLI and engine seeded workflow fixtures on local test agents so CI is deterministic

## Validation
- `pnpm --filter @smithers-orchestrator/graph build`
- `pnpm --filter @smithers-orchestrator/integrations build`
- `pnpm --filter @smithers-orchestrator/graph test`
- `pnpm --filter @smithers-orchestrator/integrations test`
- `pnpm --filter @smithers-orchestrator/graph typecheck`
- `pnpm --filter @smithers-orchestrator/integrations typecheck`
- `pnpm check:deps`
- `pnpm -r --filter "!smithers-workflows" typecheck`
- `pnpm -r typecheck`
- `pnpm lint`
- `SMITHERS_COVERAGE_PACKAGES=apps/cli pnpm coverage`
- `bun test --timeout=60000 --max-concurrency=1 packages/engine/tests/workflow-pack-smoke.e2e.test.js`
- `pnpm --filter @smithers-orchestrator/engine test`

Closes #504


roninjin10 (commented):

The core is right and valuable: one `.d.ts` per source module plus `"./*".types → "./src/*.d.ts"`, so `@smithers-orchestrator/graph/<subpath>` and `.../integrations/<subpath>` finally resolve real types instead of the barrel. I rebased onto `main`, hardened the tooling, and stripped the bundled noise.

### Added: a freshness gate (the main gap)
There was no guard against the committed declarations drifting from source — the exact failure mode `check-docs`/`check-llms` exist to prevent. I added `scripts/check-dts.mjs` (clean → `tsup` rebuild → `git status` on `**/*.d.(ts|cts|mts)`), wired into the `pnpm test` gate. Building it surfaced a real determinism bug: with no `format` pinned, tsup emits `.d.cts` in this environment, not the committed `.d.ts` — so a naive gate would be permanently red. Fixed by pinning `format: ["esm"]` in both tsup configs; regeneration is now byte-identical to the committed output (gate passes clean).

### Hardened: the entry walker
`declarationEntries()` was duplicated verbatim in both tsup configs and silently (a) overwrote on a `.js`/`.ts` twin basename collision — the exact hazard the integrations type-twins (`<name>Types.ts`) exist to avoid — and (b) skipped unexpected extensions. Hoisted it to `scripts/declaration-entries.mjs` and made it **throw** on both, with an actionable message.

### Kept (reviewer premise corrected): `ignoreDeprecations`
The suggestion to drop `ignoreDeprecations: "6.0"` from `graph/tsconfig.json` doesn't hold — removing it breaks the dts build with `TS5101: Option 'baseUrl' is deprecated`. The project sets no `baseUrl`; tsup injects one into the dts compiler options, and TS 6.0.3 flags it. I kept the option and added a comment explaining it's silencing a tooling-injected deprecation, not a project one.

### Nits
- `clean-dts.mjs` now has a header explaining why it prunes stale `.d.ts` (tsup only writes, never removes) before each build.
- Fixed the missing trailing newline on `graph/package.json`.

### Stripped
The 11 gateway-ui `@jsxImportSource react` pragmas (they're #512's — will already be on `main` when this rebases), the fuzzy-select hunk, and the three `writeFakeCodexBinary` e2e fixture removals (they belong with the codex-auth/pack-refresh work, not here). **Merge order: land #512 first**, then rebase this.

`pnpm typecheck` green; graph+integrations subpath type tests 15/15; `check-dts` passes with zero drift. Fable re-review + merge to follow (fable re-reviews briefly rate-limited).

roninjin10:
## Final re-review — verdict: NITS-ONLY (approve-worthy)

The rework is solid: per-module declarations are fresh, the freshness gate genuinely detects drift, the entry walker's skip/throw behavior is correct, and the docs row is in place. No blocking or major findings.

### What I verified (against head `d341fd011`, in an isolated worktree)

- **`node scripts/check-dts.mjs` passes** — committed per-file `.d.ts` for graph and integrations are byte-fresh, exit 0.
- **Drift detection works, empirically:**
  - Appended a type to `packages/graph/src/RetryPolicy.ts` → gate prints `DRIFT`, lists `M …/RetryPolicy.d.ts`, exits **1**, and still checks the remaining package (reports all failures, no early bail).
  - Deleted a source module (`worktree-path.js`) → the clean-then-rebuild flow surfaces the stale `.d.ts` as drift (this is exactly the case `clean-dts.mjs` exists for; tsup alone never prunes).
  - Untracked new `.d.ts` is caught too — the porcelain filter matches `??` lines.
- **Twin-collision throw works:** added `src/extract.ts` beside `extract.js` → build fails with the actionable `declarationEntries: declaration key "extract" is claimed by both extract.js and extract.ts…` message. README.md files exist at **six** levels under the two `src/` trees and are correctly skipped (builds pass).
- **Determinism:** both packages are `"type": "module"` and pin `format: ["esm"]`, so dts output is deterministically `.d.ts` (the porcelain regex also guards `.d.cts`/`.d.mts` just in case).
- **No orphans:** every committed `.d.ts` under both `src/` trees has a `.ts`/`.js` twin, so the clean → rebuild cycle is lossless.
- **`ignoreDeprecations: "6.0"` is load-bearing, verified:** graph's tsconfig has no `baseUrl` of its own — **tsup injects `baseUrl` into its dts build**, and TS 6.0.3 then fails with `error TS5101: Option 'baseUrl' is deprecated…` without the flag (plain `tsc --noEmit` passes either way). Keeping it bare is correct since a `//` comment breaks `tsconfig-checkjs.test.js`.
- **Tests:** `bun test packages/graph/tests packages/integrations/tests` → **309 pass, 0 fail**. The new/extended subpath tests are real no-mocks typechecks (symlinked `node_modules`, real `tsc`, positive + negative fixtures, deep subpaths like `utils/tree-ids` and `telegram/components/SendMessage`).
- **Docs:** `check:dts` is documented in `docs/reference/package-configuration.mdx` (both the `test` gate row and its own `check:dts` row); `node scripts/check-docs.mjs` and `node scripts/check-llms.mjs` both pass, so the llms bundles in the diff are current.
- **Scope:** diffstat confirms the gateway-ui pragma, fuzzy-select, and codex-fixture edits from the earlier revision are gone.

### MINOR / nits (none blocking)

1. **`scripts/check-dts.mjs:26` — the gate mutates tracked files in the developer's working tree.** Clean → rebuild rewrites `src/**/*.d.ts` in place during `pnpm test`; when fresh the bytes are identical so git stays clean, but on drift it leaves regenerated files dirty, and in a shared working tree (this repo's concurrent-agent setup) a gate that rewrites tracked files can race a concurrent snapshot/`git add`. A future hardening could build into a temp `outDir` and diff against committed files instead.
2. **`scripts/declaration-entries.mjs:33` — only `.js`/`.ts` count as modules.** A future `.tsx`/`.mts` under `src/` would silently ship without a declaration entry. Fails loud at the consumer (types resolution error, not silent `any`), and neither package has such files today — but a comment or a throw-on-unknown-module-extension would make the invariant explicit.
3. **`packages/graph/tsconfig.json:12` — the "why" of `ignoreDeprecations` lives only in commit history.** Since the file can't carry a comment, consider one line in `packages/graph/tsup.config.ts` (which already documents the esm pin) noting that tsup's dts build injects `baseUrl` and trips TS5101 on TS 6.
4. **`scripts/clean-dts.mjs:19` — deletes every `.d.ts` under the root, including any hypothetical hand-written one without a source twin.** None exist today (verified), and `check-dts` would flag the deletion as drift, so this fails safe. Noting for the record only.

---
*Automated fable re-review (Fable 5) — independent final pass; verified in a detached worktree at `d341fd011`, no code modified on the branch.*

PR #515, opened 2026-07-03
fix: stabilize CI gates, streams, and CLI fixtures

cookesan:
## Summary
- stop gateway event iterators as soon as caller aborts so queued frames do not leak after cancellation
- keep extension streams from yielding stale queued frames after abort
- make CLI fuzzy filtering own query edits so backspace and printable input stay deterministic in tests
- align workflow UI e2e auth and seeded responses with the isolated harness
- cover the 0.27.0 release-content tests in the .smithers package script and bring the changelog in line with docs style

## Validation
- bun test apps/cli/tests/fuzzy-select.test.js
- bun test ./.smithers/tests/release-content-probe.test.ts ./.smithers/tests/release-content-sidebar.test.ts
- node scripts/check-docs.mjs
- node scripts/check-smithers-test-script.mjs
- pnpm --filter ./.smithers test
- pnpm test
- pnpm typecheck
- pnpm lint

### DIFF
--- a/apps/cli/src/fuzzy-select.js
+++ b/apps/cli/src/fuzzy-select.js
@@ -143,16 +143,10 @@ function terminalRows() {
     return process.stdout.rows || 24;
 }
 
-function isPrintableQueryInput(seq, key) {
-    if (typeof seq !== "string" || seq.length !== 1) return false;
-    if (key?.ctrl || key?.meta) return false;
-    if (key?.name && key.name.length > 1 && key.name !== "space") return false;
-    return seq >= " ";
-}
-
 /**
  * A type-to-filter picker built on clack's `Prompt` base class. We subclass
- * `Prompt` directly (NOT `SelectPrompt`) and own the typed query so letters
+ * `Prompt` directly (NOT `SelectPrompt`) with `trackValue=true` so the base
+ * pipes readline and keeps `this.value` in sync with the typed query — letters
  * (including j/k/h/l) become filter text, while arrow keys still move the cursor.
  *
  * Instance fields we own (separate from base `this.value`/`this._cursor`):
@@ -160,8 +154,12 @@ function isPrintableQueryInput(seq, key) {
  *  - `allOptions`   immutable input options.
  *  - `filtered`     current fuzzyFilter(query, allOptions) result.
  *  - `optionCursor` index into `filtered` (clamped), NOT into allOptions.
- *      We deliberately do NOT reuse the base's `_cursor`; this keeps the
- *      highlight independent of readline's text cursor.
+ *      We deliberately do NOT reuse the base's `_cursor`: @clack/core's
+ *      `Prompt.onKeypress` runs `this._cursor = this.rl?.cursor ?? 0` on every
+ *      tracked keypress (before it emits our "cursor" event), so the base's
+ *      readline text cursor would clobber our option index — arrow-down would
+ *      stick near the top and Enter could return a lower-ranked match. Owning
+ *      `optionCursor` keeps the highlight independent of readline.
  *  - `maxItems`     windowing budget.
  *
  * Enter, ctrl-c and escape flow to the base: Enter sets state="submit" and we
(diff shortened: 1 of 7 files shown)


roninjin10 (commented):

The core `events(signal)` abort fix is real and I kept it; I rebased onto `main`, tightened it, and dropped two unrelated bundled changes.

### Kept + tightened: the queued-frame leak on abort
`SmithersGatewayConnection.events(signal)` closed the connection on abort but the drain loop condition (`while (!this.closed || this.queue.length > 0)`) keeps running while the queue still holds frames — so events that arrived *before* the abort were still yielded *after* it. Re-checking the signal on both sides of `await this.shift()` is the right fix. I removed the redundant local `aborted` flag (it's always equal to `signal.aborted`, which the listener sets synchronously) and added a comment explaining why `close()` alone is insufficient. The two tests (the discriminating one in `gateway-client.test.ts` asserting `{done:true}` with a frame queued before abort, and the tightened `extension.test.ts` asserting exactly `[{line:"first"}]`) both pass.

### Dropped
- **The `fuzzy-select.js` `trackValue=false` rewrite.** It regresses real behavior (paste/emoji input dropped, fragile vim-alias/cursor ordering) and the "bug" it targets doesn't reproduce under the locked `@clack/core@1.4.2`. Reverted to `main`.
- **The `docs/changelogs/0.27.0.mdx` churn.** 0.27.0 is a frozen release changelog; restyling it isn't in scope for this PR and no gate needs it. Reverted to `main`.

### Kept as-is (verified sound)
- `workflow-ui-all.e2e.test.js` fixture + failure-detail logging, and the `e2e-helpers.js` fake-agent response fields. Note: the codex `auth.json` chatgpt-mode fixture is coupled to the codex auth-gating in #512 — I'll land #512 before this so the two stay consistent. (This e2e is CI-skipped — no agent CLIs on the runner.)

`pnpm typecheck` green; gateway-client suites 26/26. Fable subagent re-review + merge to follow (fable re-reviews are briefly rate-limited; I'll complete them before merging).

roninjin10:
**Verdict: LGTM — the queued-frame abort leak fix is correct, tests discriminate, and the unrelated churn is gone.**

## Blocking / Major

None.

## Minor / nits

- `packages/gateway-client/tests/extension.test.ts:249` — the tightened `expect(out).toEqual([{ line: "first" }])` also passes against pre-fix code: `streamExtension` yields `initial` from the RPC response and only then enters `connection.events(signal)`, where the pre-existing early `signal?.aborted` check already stops iteration. So this test alone is not the regression guard — the new `gateway-client.test.ts` test is (verified: it fails against `origin/main` source). Fine as-is since both land together; noting so nobody later "simplifies" away the connection-level test thinking the extension test covers it.

## What I verified

- **Fix correctness** (`packages/gateway-client/src/SmithersGatewayConnection.ts:113-127`): re-checking `signal.aborted` before and after `await this.shift()` means no frame is yielded after abort — including the race where a frame is queued while `shift()` is pending and abort fires before the continuation runs (a frame may be shifted and discarded post-abort, which is correct since the abort handler closed the connection).
- **Clean close still drains**: the `while (!this.closed || this.queue.length > 0)` condition is untouched; without an aborted signal the loop drains queued frames then ends on the `close` sentinel.
- **No hang / no listener leak**: abort calls `close()`, which pushes a `close` item that wakes any pending `shift()` waiter; the abort listener is `{ once: true }` and removed in `finally`; the pre-loop already-aborted path returns before `addEventListener`.
- **Callers**: all three production consumers of `events(signal)` (`streamRunEvents`, `streamDevTools`, `streamExtension` in `SmithersGatewayClient.ts`) are subscription streams where abort means caller cancellation — none relied on the old drain-after-abort behavior.
- **Tests run at head `b12e871a4`**: `bun test packages/gateway-client/tests/gateway-client.test.ts packages/gateway-client/tests/extension.test.ts` → 26 pass, 0 fail. With only `SmithersGatewayConnection.ts` reverted to `origin/main`, the new "abort stops the event stream without draining queued frames" test fails (Expected -1 / Received +0) — it discriminates. The extension test's exact `[{line:"first"}]` assertion is deterministic: both the initial RPC response and the "live" event frame are dispatched in the same microtask, so "live" is guaranteed queued before the consumer aborts.
- **Typecheck**: `tsc --noEmit -p packages/gateway-client` → exit 0.
- **Churn dropped**: net diff `origin/main...b12e871a4` is exactly 5 files (the fix, the two gateway-client tests, and the two CLI test-fixture tweaks in `apps/cli/tests/workflow-ui-all.e2e.test.js` / `packages/smithers/tests/e2e-helpers.js`). No fuzzy-select rewrite, no `docs/changelogs/0.27.0.mdx` churn.
- **Fixture tweaks**: test-only — realistic fake `.codex/auth.json` (chatgpt auth mode) with `OPENAI_API_KEY=""`, better run-failure diagnostics (last 20 output lines), and two added fields (`polished`, `changesMade`) in `FAKE_AGENT_RESPONSE`. All consistent with the seeded-fake-agent CI convention.

---
*Automated fable re-review (final pass) of head `b12e871a4`.*

PR #516, opened 2026-07-03
fix(db): surface mounted approvals

cookesan:
## Summary
- include waiting approval nodes in pending-approval queries when a request row is missing
- keep the fallback scoped to active waiting-approval runs
- add SQLite and Postgres coverage for the missing-row case
- fix a CLI fuzzy-picker backspace path that left control bytes in the filter text
- keep workflow UI e2e fixtures offline with the existing subscription-auth fixture
- include existing release-content tests in the workflow-pack script and normalize docs that current checks enforce

## Tests
- bun test packages/db/tests/db-adapter.test.js
- bun test packages/db/tests/db-postgres-dialect.test.js
- pnpm --filter @smithers-orchestrator/db test
- pnpm --filter @smithers-orchestrator/db typecheck
- bun test packages/server/tests/gateway.test.jsx packages/gateway-client/tests/data/electric-row-parity.test.ts apps/cli/tests/tui-unit.test.js
- node scripts/check-docs.mjs
- pnpm --filter smithers-workflows test
- node scripts/check-smithers-test-script.mjs
- node scripts/check-llms.mjs
- bun test apps/cli/tests/fuzzy-select.test.js
- bun test apps/cli/tests/workflow-ui-all.e2e.test.js
- pnpm --filter @smithers-orchestrator/cli test
- pnpm typecheck
- pnpm lint
- pnpm test

### DIFF
--- a/apps/cli/src/fuzzy-select.js
+++ b/apps/cli/src/fuzzy-select.js
@@ -138,6 +138,18 @@ function limitOptions({ cursor, options, maxItems, rows, style }) {
     });
 }
 
+function applyBackspaceControls(input) {
+    let output = "";
+    for (const char of String(input ?? "")) {
+        if (char === "\x7f" || char === "\b") {
+            output = output.slice(0, -1);
+            continue;
+        }
+        output += char;
+    }
+    return output;
+}
+
 /** process.stdout.rows is undefined when not a TTY (CI/piped); fall back to 24. */
 function terminalRows() {
     return process.stdout.rows || 24;
@@ -207,6 +219,17 @@ class FuzzySelectPrompt extends Prompt {
             if (this.optionCursor > this.filtered.length - 1) {
                 this.optionCursor = Math.max(0, this.filtered.length - 1);
             }
+        });
+
+        this.on("key", (_seq, key) => {
+            if (key?.name !== "backspace" && key?.name !== "delete") return;
+            const normalized = applyBackspaceControls(this.userInput);
+            if (normalized === this.userInput) return;
+            if (this.rl) {
+                this.rl.line = normalized;
+                this.rl.cursor = Math.min(this.rl.cursor ?? normalized.length, normalized.length);
+            }
+            this._setUserInput(normalized);
         });
 
         // Arrow keys (and only arrow keys, since trackValue=true) emit "cursor".
--- a/apps/cli/src/fuzzy-select.js
+++ b/apps/cli/src/fuzzy-select.js
@@ -138,18 +138,6 @@ function limitOptions({ cursor, options, maxItems, rows, style }) {
     });
 }
 
-function applyBackspaceControls(input) {
-    let output = "";
-    for (const char of String(input ?? "")) {
-        if (char === "\x7f" || char === "\b") {
-            output = output.slice(0, -1);
-            continue;
-        }
-        output += char;
-    }
-    return output;
-}
-
 /** process.stdout.rows is undefined when not a TTY (CI/piped); fall back to 24. */
 function terminalRows() {
     return process.stdout.rows || 24;
@@ -219,17 +207,6 @@ class FuzzySelectPrompt extends Prompt {
             if (this.optionCursor > this.filtered.length - 1) {
                 this.optionCursor = Math.max(0, this.filtered.length - 1);
             }
-        });
-
-        this.on("key", (_seq, key) => {
-            if (key?.name !== "backspace" && key?.name !== "delete") return;
-            const normalized = applyBackspaceControls(this.userInput);
-            if (normalized === this.userInput) return;
-            if (this.rl) {
-                this.rl.line = normalized;
-                this.rl.cursor = Math.min(this.rl.cursor ?? normalized.length, normalized.length);
-            }
-            this._setUserInput(normalized);
         });
 
         // Arrow keys (and only arrow keys, since trackValue=true) emit "cursor".
--- a/apps/cli/tests/fuzzy-select.test.js
+++ b/apps/cli/tests/fuzzy-select.test.js
@@ -145,12 +145,8 @@ class FakeInput extends EventEmitter {
     on(event, cb) {
         return super.on(event, cb);
     }
-    /** Drive the same readline path a real TTY uses, including line editing. */
+    /** Drive the prompt's onKeypress directly, mirroring readline keypress events. */
     keypress(seq, key) {
-        if (typeof seq === "string") {
-            this.emit("data", Buffer.from(seq));
-            return;
-        }
         this.emit("keypress", seq, key);
     }
 }
--- a/apps/cli/tests/fuzzy-select.test.js
+++ b/apps/cli/tests/fuzzy-select.test.js
@@ -145,8 +145,12 @@ class FakeInput extends EventEmitter {
     on(event, cb) {
         return super.on(event, cb);
     }
-    /** Drive the prompt's onKeypress directly, mirroring readline keypress events. */
+    /** Drive the same readline path a real TTY uses, including line editing. */
     keypress(seq, key) {
+        if (typeof seq === "string") {
+            this.emit("data", Buffer.from(seq));
+            return;
+        }
         this.emit("keypress", seq, key);
     }
 }
--- a/apps/cli/tests/workflow-ui-all.e2e.test.js
+++ b/apps/cli/tests/workflow-ui-all.e2e.test.js
@@ -132,19 +132,12 @@ workflowUiTest("every init-pack workflow UI builds + boots; the output-verified 
     HOME: repo.dir,
     PATH: [binDir, "/usr/bin", "/bin", "/usr/sbin", "/sbin"].join(delimiter),
     ANTHROPIC_API_KEY: "",
-    OPENAI_API_KEY: "",
+    OPENAI_API_KEY: "sk-test-openai-key",
     GEMINI_API_KEY: "",
     GOOGLE_API_KEY: "",
   };
   repo.write(".claude/.credentials.json", "{}\n");
-  repo.write(
-    ".codex/auth.json",
-    JSON.stringify({
-      auth_mode: "chatgpt",
-      OPENAI_API_KEY: null,
-      tokens: { access_token: "fake-access-token", account_id: "acct_test" },
-    }) + "\n",
-  );
+  repo.write(".codex/auth.json", "{}\n");
   repo.write(".gemini/antigravity-cli/settings.json", "{}\n");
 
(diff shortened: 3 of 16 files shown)


roninjin10 (commented):

The fallback that surfaces waiting-approval nodes with no `_smithers_approvals` row is valid, correctly scoped (anti-join `a.run_id IS NULL`, run gated to `waiting-approval`), and dialect-parity-checked. I rebased onto `main`, added the root-cause fix the fallback was papering over, and stripped two things that don't belong.

### Added (MAJOR): fix the root cause, not just the symptom
The reason `_smithers_approvals` was empty for detached runs is that `buildMirrorOnProgress` (`packages/server/src/index.js`) never projected the approval events — it handled `NodeWaitingApproval` but not `ApprovalRequested`/`Granted`/`AutoApproved`/`Denied`. So the mirror had node state but no approval rows, and the fallback derived **bare gates** from node state — dropping `request_json`, so select/rank/decision approvals rendered as plain gates.

I added the projection (mirroring the engine's own `approveNode`/deferred-state-bridge writes): `ApprovalRequested` reconstructs the row with `request_json`; the decision events flip it to `approved`/`denied` (preserving the request payload, since those events carry no decision detail) so it stops surfacing as pending. The fallback stays as defense-in-depth.

### Tests
- Kept the fallback-positive test (waiting node with no row surfaces; a cancelled run's waiting node does not).
- Added an **anti-join negative** test: a node still reading `waiting-approval` but with a *decided* approval row must NOT be resurfaced as pending (guards the granted-approval-reappears regression).
- Added a **mixed** test: a real `requested` row (with payload) and a fallback row coexist without duplication, and only the real one keeps its `request_json`.

### Stripped
- The `compute-task-bridge.js` change that dropped `{ signal: taskSignal }` from the supervisor race — that removes external cancellation on the watchdog path (a run cancel would hang), and it conflicts with #521. Reverted to `main`.
- The bundled `fuzzy-select.js` hunk (no real bug under locked `@clack/core`).

### Kept as-is
The hot-controller/overlay changes + their tests, and the `timer.test.jsx` timeout bump (stability). `pnpm typecheck` green; db-adapter 56/56, server 39/39, postgres-dialect 16/16, hot-controller suites 33/33.

Fable re-review + merge to follow (fable re-reviews briefly rate-limited).

roninjin10:
## Final re-review of 61457d11e

**Verdict: NEEDS-CHANGES** — the mirror projection and DB fallback are correct and the previously flagged regressions are confirmed reverted, but the root-cause fix itself (the approval-event projection) ships with zero test coverage.

### MAJOR

- **`packages/server/src/index.js:719-763` — the new `ApprovalRequested`/`ApprovalGranted`/`ApprovalAutoApproved`/`ApprovalDenied` projection in `buildMirrorOnProgress` is untested.** The new db tests cover only the fallback layer (anti-join, mixed rows, pg dialect); `packages/server/tests/server.test.js`'s approve/deny tests hit the run's own DB, not the mirror, and only `mapEvent.test.js` mentions these event types. If the projection regresses, `_smithers_approvals` in the mirror goes empty again and every test stays green — the fallback silently downgrades select/rank/decision approvals to bare gates, which is exactly the bug this PR fixes. One test that feeds an ApprovalRequested then a decision event through the mirror path (e.g. a detached run against `startTestServer({ db })`, asserting `request_json` survives into the pending list and the row leaves pending after the decision) would lock in the PR's central claim.

### MINOR / nits

- `packages/server/src/index.js:772-780` — mirror events are dispatched fire-and-forget (`void runPromise(...)` per event), so the decision handler's read-modify-write (`getApproval` at :741 then upsert) can theoretically interleave with a still-in-flight `ApprovalRequested` projection and flip a decided mirror row back to `requested`. In practice the two events are separated by human decision latency (and auto-approve emits no `ApprovalRequested`), and the fire-and-forget pattern predates this PR — noting for the record only.
- `packages/db/src/adapter.js:2630-2647` — the fallback branch requires `r.status = 'waiting-approval'`, so a mirror-lagged run still reporting `running` with a waiting node gets no fallback gate. Fine now that the projection creates real rows; just confirming this narrowing is intentional (it keeps stale gates in finished/cancelled runs from resurfacing).
- Scope: the PR still carries the first commit's unrelated engine hot-reload changes (`packages/engine/src/hot/overlay.js` copy-instead-of-hardlink for module/JSON files, `HotWorkflowController.closeSync`) and test deflakes (`packages/components/tests/timer.test.jsx`, `packages/engine/tests/engine.test.jsx`) under a db-fix title. All are individually sensible and green, but this bundling is what the repo's atomic-commit norm exists to prevent — history note, not a code issue.

### Verified

- **Projection correctness:** decision events carry only ids + timestamp (`packages/engine/src/approvals.js:162-168`), so preserving `requestJson`/`note`/`decidedBy`/`decisionJson` from the existing row is right; `autoApproved` is set true only for `ApprovalAutoApproved`. `ApprovalRequested` carries `request` (`packages/engine/src/effect/deferred-state-bridge.js:1158`) and the reconstructed row matches the source-of-truth row shape. Rows come back camelCased (`snakeToCamel`), so `existing?.requestedAtMs` etc. are real fields.
- **Fallback:** the anti-join (`a.run_id IS NULL`) excludes nodes with *any* approval row, so decided approvals never resurface — covered by the new "never resurfaced" test. The explicit 11-column SELECT list equals the full `_smithers_approvals` schema (lossless vs the old `SELECT *`). `auto_approved` is INTEGER in both dialects, so `0 AS auto_approved` unions cleanly; the pg-dialect test exercises the fallback on the postgres-dialect storage.
- **Stripped hunks:** `packages/engine/src/effect/compute-task-bridge.js`, `apps/cli/src/fuzzy-select.js`, and their tests are byte-identical to the merge-base at PR head (the small two-dot delta vs main tip is main-side drift from b3d4252df).
- **Ran at PR head** (detached worktree): `db-adapter.test.js` 56 pass, `db-postgres-dialect.test.js` 16 pass, `server.test.js` 39 pass, engine hot-controller/races/overlay + engine.test.jsx 46 pass, components `timer.test.jsx` 8 pass, root `tsc --noEmit` clean.

---
*Automated fable re-review (claude-fable-5), final pass over `fix/approval-node-fallback` @ 61457d11e.*

roninjin10:
## Re-review (follow-up): NEEDS-CHANGES finding resolved

**Verdict: RESOLVED (nits only)** at head `916ea73c5`.

The single blocking finding from the prior re-review — the ApprovalRequested/Granted/AutoApproved/Denied mirror projection in `buildMirrorOnProgress` (`packages/server/src/index.js`) was untested — is now closed by `packages/server/tests/mirror-approval-projection.test.js`.

### Why the new test is adequate

**It genuinely exercises the projection, not the node-state fallback.** The fallback UNION branch in `listPendingApprovals` (`packages/db/src/adapter.js`) (a) always emits `NULL AS request_json` and (b) requires an `INNER JOIN _smithers_runs` row with `status = 'waiting-approval'`. The test never inserts a run row and `NodeWaitingApproval` in the mirror only writes node state, so the fallback branch is structurally dead here — the `requestJson != null` + `"decision"`/`"ship"` content assertions can only be satisfied by the projected `_smithers_approvals` row.

**It is discriminating — mutation-verified.** In a scratch worktree at `916ea73c5`:
- Disabling the `case "ApprovalRequested"` projection → test fails at the pending-row assertion (test line 57).
- Disabling the `case "ApprovalGranted"` projection → test fails at the leaves-pending assertion (test line 67).
- Unmutated: test passes (1 pass, 6 assertions).

**Gates checked at the PR head:** full `packages/server` suite green (388 pass, 1 skip, 0 fail) and root `pnpm typecheck` clean, so the new `__serverTestInternals` export breaks nothing.

### Non-blocking nits (no action required)

1. The `ApprovalDenied` / `ApprovalAutoApproved` arms of the shared case block aren't directly asserted (`status: "denied"`, `autoApproved: true`). They share the tested preserve-existing-row logic, so risk is low.
2. `__serverTestInternals` is the first test-internals export in `packages/*/src`. It's dunder-named, commented, and absent from `index.d.ts` (bare-specifier consumers won't see it), which is acceptable — extracting `buildMirrorOnProgress` to its own module would be the cleaner long-term shape if more internals accrue.

---
*Automated fable re-review (claude-fable-5); scope limited to verifying the previously flagged untested-projection gap.*

roninjin10 (changes requested):
Re-reviewed after the rebase (head 916ea73c5). The core of this PR is now clearly justified and we want it: on main, `buildMirrorOnProgress` in `packages/server/src/index.js` projects `NodeWaitingApproval` into the server DB but has no `ApprovalRequested`/`ApprovalGranted`/`ApprovalDenied` cases, so approvals on detached (mirrored) runs never get a `_smithers_approvals` row and are invisible to `/v1/approvals`, MCP `list_pending_approvals`, and `smithers ps`. Your mirror projection fix + the db-layer fallback + `mirror-approval-projection.test.js` (which exercises the real projector against a real sqlite DB) are merge-worthy as-is.

Four changes needed before merge:

1. **Drop the `apps/cli/tests/workflow-ui-all.e2e.test.js` hunk — it reverts #515.** Main (via your own #515, merged minutes before this rebase) now uses the chatgpt subscription-auth `.codex/auth.json` fixture with `OPENAI_API_KEY: ""`. This PR puts back `sk-test-openai-key` + empty `auth.json`, the opposite of what the PR body says it does.

2. **Drop the `packages/components/tests/timer.test.jsx` timeout bumps — mooted by #517.** With timer anchors carried across continuations (merged), the file passes on current main as-is (verified: 8 pass / 0 fail, 7.3s). The bump just slows the suite.

3. **Split the hot-reload changes into their own PR** (`HotWorkflowController.js` sync close, `overlay.js` hardlink→copy, and the four engine test rewrites). The overlay copy-per-generation rationale is plausible and the hot-overlay test is good, but none of it is "fix(db): surface mounted approvals", and switching every hot-controller test to an injected `useNoopWatcher` reduces real-watcher coverage — that trade needs its own review. Same for the `engine.test.jsx` parallel-concurrency rewrite (it looks like a genuine deflake improvement — worth its own small PR where it can be evaluated as such).

4. **Rewrite the PR description** to match the actual diff — it currently mentions a fuzzy-picker fix and workflow-pack change that aren't here, and omits the server mirror projection which is the best part of the PR.

Two smaller notes, non-blocking: mirror event projection is fire-and-forget per event (`void runPromise`), so a late `ApprovalRequested` racing a decision projection could last-write-wins a decided row back to `requested` — worth a follow-up guard (e.g. skip insert if a decided row exists). And the `SELECT *` → explicit column list in `listPendingApprovals` means future `_smithers_approvals` columns won't surface until someone remembers to add them there.

With 1, 2, and 3 done, this merges immediately.

Issue #502, opened 2026-07-03
🐛 scheduler: duration timer deadline drifts (recomputed each decide() instead of anchored)


roninjin10:
## Summary
The scheduler recomputes a duration timer's deadline against the current clock on every `decide()` instead of anchoring it at the timer's start, so the fire time drifts later each time `decide()` runs during sibling work — a 5-minute timer re-evaluated at minute 3 gets pushed to minute 8.

## Location
`packages/scheduler/src/makeWorkflowSession.js` — `timerResumeAtMs()` returns `nowMs + ms` for a `__timerDuration`, and `nowMs` is the current clock at each decide.

## Impact
Duration timers never fire on time under contention; they keep postponing. (Absolute `__timerUntil` timers are already anchored via `Date.parse`, so only relative durations drift.)

## Suggested fix
Anchor the deadline to when the timer first entered `waiting-timer`. There is currently no persisted start time, so this needs either:
- the engine's `deferred-state-bridge` to record `__timerStartedAtMs` when a duration timer parks (and convert duration → absolute fire time once), or
- the scheduler to record the first-observed `waiting-timer` timestamp per task (like `quotaResetTimes`) — but that must survive resume, so a persisted anchor is preferable.

## Why deferred
The clean fix touches the engine's durable timer state (cross-package) and I wouldn't re-derive that blind. The related parser/busy-loop half of this finding (`d`-unit, case-insensitivity) was fixed in `603beae7` (fix(scheduler): parse d-unit and case-insensitive timer durations).

_From the full-codebase review; the fixable findings were landed on `main`, this one was deferred._

roninjin10:
## Complexity triage: **Tier 2 — Fable plans & reviews, Codex implements (tricky but well-scoped)**

Needs a persisted timer anchor that survives resume — either the engine's deferred-state-bridge records the start when a duration timer parks, or the scheduler persists first-observed waiting-timer time. Cross-package durable-state design; part of the timer family with #545/#538/#494, best planned as one coherent fix.

---
<sub>Triage legend — **Tier 1 (Opus)**: straightforward, well-localized, diagnosis/fix already known; Opus can one-shot it. **Tier 2 (Fable)**: requires architecture/design decisions but well-scoped; Fable plans and reviews, Codex implements. **Tier 3 (Smithers workflow)**: needs breaking down into separate tasks by Fable, run as a smithers script. Triaged 2026-07-07.</sub>

roninjin10:
[...]

PR #517, opened 2026-07-03
fix(scheduler): carry timer anchors across continuations

cookesan:
## Summary
- carry live duration-timer start anchors in scheduler continuation payloads
- restore carried timer anchors when the engine starts a continued run
- cover scheduler export/import and engine payload parsing helpers

## Why
Duration timers are anchored when the scheduler first parks them. If a continuation starts a fresh scheduler session without that anchor, the relative deadline can be recomputed from the new run start and extend the wait. This covers the continuation handoff side of #502.

## Verification
- bun test packages/scheduler/tests/workflowSession-service-branches.test.js
- bun test packages/engine/tests/engine-internals.test.js
- pnpm --filter @smithers-orchestrator/scheduler typecheck
- pnpm --filter @smithers-orchestrator/engine typecheck
- pnpm --filter @smithers-orchestrator/scheduler test
- pnpm --filter @smithers-orchestrator/engine test
- pnpm check:effect
- pnpm check:deps
- pnpm lint
- pnpm -r typecheck
- pnpm check:docs
- pnpm check:llms

### DIFF
--- a/packages/engine/src/engine.js
+++ b/packages/engine/src/engine.js
@@ -1484,6 +1484,7 @@ async function continueRunAsNew(params) {
         continueAsNewEvery: continuation.continueAsNewEvery ?? null,
         payload: continuation.statePayload ?? null,
         ralph: ralphStateToObject(carriedRalphState),
+        ...(continuation.nextTimerStarts ? { timerStarts: continuation.nextTimerStarts } : {}),
         timestampMs: ts,
     };
     const carriedStateJson = JSON.stringify(continuationEnvelope);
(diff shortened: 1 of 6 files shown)


roninjin10 (commented):

Thanks for chasing this down. The core diagnosis is right and valuable: duration-timer anchors were being dropped across a continue-as-new boundary, so every long-running timer restarted from zero after a handoff. I rebased onto `main` (the branch was 34 commits behind) and reworked the carry mechanism to close a latent hole. Summary of what changed from your version:

### Fixed (MAJOR): anchors were still clobberable by user continuation state
Your version carried `timerStarts` inside the continue-as-new **`statePayload`**. On the *explicit* `<ContinueAsNew state={...}>` path the engine JSON-parses the user's `stateJson` **over** `statePayload` (`engine.js`, `continueAsNew` handler) — so a workflow that passes its own continuation `state` would silently drop the timer anchors, which is exactly the bug this PR set out to fix, just on a different trigger.

The repo already solved this shape for ralph state: it rides a **dedicated** `ralph` envelope field and is re-extracted straight from the transition via `nextRalphState`, never through the overwritable `payload`. I made `timerStarts` follow the same pattern:

- Scheduler emits a dedicated `transition.timerStarts` field (both the explicit and loop-threshold continuation paths), no longer folded into `statePayload`/`ralphStatePayload`.
- Engine extracts it via `timerStartsFromDriverTransition` (immune to `stateJson`) and carries it on a dedicated `continuationEnvelope.timerStarts` field next to `ralph`.
- Next run reads anchors from that dedicated field (`continuationEnvelopeFromInput`/`FromConfig` + `timerStartsFromContinuationEnvelope`), not from `payload`.

### Nits applied
- Tightened numeric coercion from `Number(value)` to `typeof value === "number" && Number.isFinite(value)` on both read paths, so a stringified `"1000"` is rejected rather than silently coerced.
- Removed the now-unused `schedulerStatePayload()` indirection.

### Tests
- Reworked the scheduler test to assert the anchor lands on the dedicated `timerStarts` field and that `statePayload` stays `undefined`.
- Added an engine-level **clobber-immunity handoff test**: an explicit continuation with a populated user `stateJson` (including a spoofed `payload.timerStarts`) still recovers the real anchors through the exact envelope shape the engine writes, via both the input-row and run-config carry paths.

`pnpm typecheck` and the engine/scheduler suites are green. Dropping a fable subagent re-review next; merging once it's clean.

roninjin10:
## Re-review of `fix/timer-start-continuation` @ `988f0c7`

**Verdict: correct and mergeable — no blocking issues.** The anchor carry is sound on both continuation paths, both carry mediums, and both DB dialects. One notable follow-up (stale generated scheduler types) and a few nits below.

### Verified correct

- **Full round trip traced.** Session `timerStartsField()` → `transition.timerStarts` on both `ContinueAsNew` sites (`packages/scheduler/src/makeWorkflowSession.js:673` explicit, `:825` loop-threshold) → engine `timerStartsFromDriverTransition` (`packages/engine/src/engine.js:6167`) → `continuation.nextTimerStarts` → envelope `timerStarts` as a sibling of `payload` (`engine.js:1487`) → carried input row `payload.__smithersContinuation` **and** `config.continuation` → next run reads input-envelope-first with config fallback (`engine.js:6024-6027`) → `initialTimerStarts` seeds `state.timerStarts` → `timerResumeAtMs` re-anchors the deadline. Keys are `nodeId::iteration`, stable across the handoff.
- **Postgres parity holds.** The envelope, `inputRow`, and `nextConfigJson` are all built once in `continueRunAsNew` *before* the dialect branch (`engine.js:1479-1524`); `continueRunAsNewPostgres` receives them pre-built (`engine.js:1535-1551`) and inserts them verbatim. The new field cannot be dropped on the postgres path.
- **Clobber immunity.** User `stateJson` is parsed into `statePayload` → envelope `payload`; the reader only looks at the envelope-level `timerStarts` key, so a spoofed `payload.timerStarts` is ignored. `buildCarriedInputRow` overwrites any pre-existing `__smithersContinuation` key in the user payload (`engine.js:1323`).
- **No ralph regression.** The explicit path never set `statePayload` (before or after this PR), so `nextRalphState` stays `undefined` and `continueRunAsNew` falls back to the engine-tracked `ralphState` exactly as on `main` (`engine.js:1476-1478`). Ralph state still rides the durable `_smithers_ralph` table, untouched.
- **Coercion/edges.** `typeof === "number"` + `Number.isFinite` filtering on both write and read sides rejects numeric strings, NaN, and ±Infinity; empty keys are dropped; arrays are rejected; empty map → `undefined` so the field is omitted from the envelope/JSON; the `initialTimerStarts` seed re-filters non-finite values defensively.
- **Config fallback is durable.** `buildDurabilityConfig` spreads the existing config through (`engine.js:1765-1773`, `:5803-5820`), so `continuation` (and its `timerStarts`) survives config rewrites, including the workflow-versioning `persist` path.
- **First-render safety.** The `pruneUnmounted` sweep iterates `state.states` only (`makeWorkflowSession.js:432-448`), so seeded anchors survive graph submits until the timer actually mounts.
- **Tests are meaningful, not tautological.** The scheduler test drives real `decide()` passes, asserts `statePayload` stays `undefined` on the explicit path, and proves the re-anchored deadline (`resumeAtMs 6_000` from a session created at `now=4_000` — a fresh anchor would give `9_000`). The engine clobber test injects a spoofed `payload.timerStarts` and shows it is ignored on both mediums.

### MINOR (follow-up, not blocking)

1. **Stale published types** — `packages/scheduler/src/index.d.ts` (the package's `exports.types` entry) does not include `initialTimerStarts` on `WorkflowSessionOptions` (see `WorkflowSessionOptions$2`, ~line 237). Runtime is unaffected and root `pnpm typecheck` is green (the JSDoc-typed engine resolves the `.ts` source), but external TS consumers passing the new option would get an excess-property error. **Caveat discovered while verifying:** regeneration is currently broken repo-wide — `pnpm -C packages/scheduler build` (tsup `--dts-only`, `outDir: src`) is a fixed point because the existing sibling `src/index.d.ts` shadows `src/index.js` during dts compilation, so tsup re-emits the old file byte-for-byte; deleting it first exposes a real DTS build failure (`error TS5101: Option 'baseUrl' is deprecated` with tsup 8.5.1 + TS 6.0.3). That tooling breakage predates this PR, so I'm not blocking on it — but the dts pipeline deserves its own fix, after which this file should be regenerated.

### Nits

2. `engine.js:1350` / `:1456` — the JSDoc type `ContinueAsNewRequest` is referenced but defined nowhere in the repo (pre-existing), so the new `nextTimerStarts` field has no declared type. Worth adding a real typedef while touching this area.
3. `makeWorkflowSession.js:634` — `timerStartsField()` sorts keys with `localeCompare`, which is locale/ICU-sensitive. A plain `left < right ? -1 : 1` byte compare would make the envelope JSON byte-deterministic across environments. Cosmetic (key order only).
4. Anchors for timers that never re-mount in the child run are never pruned from `state.timerStarts` and get re-carried on every subsequent handoff. Harmless and bounded by `MAX_CONTINUATION_STATE_BYTES`, but a one-line comment (or pruning against the mounted graph) would prevent surprise.
5. Coverage gaps: (a) no test drives the actual `continueRunAsNew` envelope write — the engine test hand-builds the envelope, so a field-name drift at `engine.js:1487` vs the readers would stay green; (b) the ralph **loop-threshold** transition site carrying `timerStarts` is untested (code-identical to the explicit site, so low risk); (c) the postgres continue-as-new test (`create-smithers-postgres.test.jsx`) doesn't assert `timerStarts` carriage, though parity is structural as noted above.

### What I ran

- `pnpm typecheck` — green.
- `bun test packages/engine/tests/engine-internals.test.js` — 22/22 pass.
- Full `packages/scheduler` suite — 137/137 pass.
- Full `packages/engine` suite — 684/687; the 3 failures are `workflow-command.e2e.test.js` timeouts that fail identically on `origin/main` in this environment (verified in a clean worktree) — unrelated to this PR.
- Verified the dts regeneration behavior described in finding 1 first-hand (rebuilt, checked mtime + content, restored).

---
*Automated fable re-review (Claude Fable 5) of head `988f0c7a7a42cd59387c529326f6e28fe0294eaa`.*

Issue #509, opened 2026-07-03
Epic: workflow testing library (smithers-orchestrator/testing)


roninjin10:
# Epic: workflow testing library (`smithers-orchestrator/testing`)

Full design spec: `.smithers/specs/workflow-testing-library.md`

## Why

Workflows fail on their first real run for reasons that are all knowable before an agent spends a token: a prompt that renders wrong, a `deps` key mapping to a node id nothing produces (silent `DEPENDENCY_DEADLOCK`), a schema mismatch between one task's output and the next task's read, a `<Loop until>` reading `outputMaybe` instead of `latest` so it never advances, an approval gate whose condition never fires, a compute task that throws then retries forever (compute `retries` default to `Infinity`).

Today the only ways to catch these are `smithers graph` (one frame, load-time bugs only), the fake-binary CLI e2e harness (high fidelity, heavyweight, internal), `smithers eval` (spends tokens), or shipping and watching it fail. There is also no way to state an expectation the type system cannot capture ("after triage runs, more than one ticket exists") and have it enforced in tests and at runtime. Scorers observe but never gate; no assertion primitive exists.

Goal: make workflow tests fast (milliseconds, no DB or subprocess for the common tiers), typed (mock values validated against each task's real output table before anything runs), with an assertion primitive that runs in tests and in production alike. The user asked for a vitest library specifically; the core is runner-agnostic and vitest-compatible, with `bun test` as the fallback for the full engine tier.

## Shape: three tiers, one mock vocabulary

```
tier 1  renderWorkflow()  render one frame, no execution        Node + Bun
tier 2  simulate()        full control flow, in-memory, mocked  Node + Bun   <- the workhorse
tier 3  testWorkflow()    real engine, temp DB, real files      Bun only
        e2e (existing)    real CLI + fake binaries              re-exported
```

New package `packages/testing` (`@smithers-orchestrator/testing`), source-shipping TS (no dts build), subpaths `/testing`, `/testing/engine`, `/testing/e2e`, `/testing/vitest`.

```ts
import { simulate, fakeAgent, auto } from "smithers-orchestrator/testing";
import workflow from "../workflows/hello";

const sim = simulate(workflow, {
  input: { name: "Ada" },
  mocks: { greet: { greeting: "Hi Ada" }, "*": auto }, // bare value validated vs the node's real schema
});
await sim.run();
expect(sim.status).toBe("finished");
expect(sim).toHaveExecutedInOrder(["greet", "output"]);
[...]

PR #518, opened 2026-07-03
feat(testing): add workflow test helpers

cookesan:
## Summary
- add `@smithers-orchestrator/testing` with `fakeAgent`, `auto`, `renderWorkflow`, `renderPrompt`, and `runTask`
- expose the helpers through `smithers-orchestrator/testing`
- cover schema validation, scripted fake agents, file writes, frame rendering, prompt rendering, and descriptor task execution
- update package docs and generated llms artifacts

Part of #509.

## Tests
- `pnpm --filter @smithers-orchestrator/testing typecheck`
- `pnpm --filter @smithers-orchestrator/testing test`
- `pnpm --filter smithers-orchestrator typecheck`
- `pnpm -r typecheck`
- `pnpm lint`
- `pnpm check:effect`
- `pnpm check:deps`
- `pnpm check:docs`
- `pnpm check:llms`
- `pnpm typecheck:examples`
- `pnpm typecheck:evals`
- `bun -e 'import { fakeAgent, renderWorkflow, runTask } from "smithers-orchestrator/testing"; if (typeof fakeAgent !== "function" || typeof renderWorkflow !== "function" || typeof runTask !== "function") process.exit(1); console.log("testing subpath ok")'`
- `bun -e 'import { fakeAgent } from "@smithers-orchestrator/testing"; if (typeof fakeAgent !== "function") process.exit(1); console.log("testing package ok")'`


roninjin10 (commented):

Solid foundation for the testing library. I rebased onto `main` and fixed the two correctness issues; below I'm also candid about where the Node-compat goal actually stands, because it's bigger than it looks.

### Fixed (MAJOR): renderPrompt must use the production renderer
`renderPrompt` hand-copied the prompt→text logic (entity decode, markdown-component injection, blank-line collapse) and had already started to diverge from `packages/components`' `renderPromptToText` — so a test could pass while the engine renders the prompt differently. I exported `renderPromptToText` from the components barrel (the "new export") and re-export it as `renderPrompt`, so there is exactly one implementation.

### Fixed (MAJOR): fakeAgent wrapper detection
`normalizeResult` treated any value with `output`/`text`/`files` keys as the `{output,text,files}` wrapper — so a bare task output whose own schema has those field names was mis-parsed. It now tries validating the value against the output schema first (if it validates, it IS the output) before falling back to the wrapper shape.

### Fixed: publish gate + docs
Bumped `@smithers-orchestrator/testing` 0.26.1 → 0.28.0; reverted the stale v0.26.1 llms regen and regenerated the current bundles at 0.28.0.

### The Node-compat decision — what I found (please read)
The goal is Node/vitest support, and the package ships raw `.ts`. I built the compile path end-to-end and hit three real blockers that make "just compile it" a larger, separate effort, not a rider on this PR:
1. **`dist/` is globally gitignored** — compiled output can't ship there. The repo's convention is to commit `.js`+`.d.ts` into `src/`, which for a `.ts`-authored package means a full hand-port of all five modules (incl. the JSX `renderWorkflow` and the generic `fakeAgent`).
2. **Node runtime is blocked deeper than this package**: even after compiling to `.js`, `import`-ing `@smithers-orchestrator/testing` under Node throws `ERR_UNSUPPORTED_ESM_URL_SCHEME: Received protocol 'bun:'` — the dependency chain (components → graph → …) transitively imports `bun:` builtins. Real Node support needs those upstream packages to be `bun:`-free first.
3. Consequently a "Node import-clean" CI gate would be **red today**, so I did not add one.

My recommendation: keep this PR Bun-shipped with the correctness fixes, and track the Node-compat work as its own effort (hand-port `.ts`→`.js`+`.d.ts` in `src/` for this package + make the transitive dep chain `bun:`-free + then add the import-clean gate).

### Spec deviations — accepting, with deltas to note on the epic
Per decision, keeping the shipped API where it's coherent rather than forcing it to the #509 spec: `runTask`'s signature, `renderPrompt` being sync, and `auto` being a plain object (not callable). Worth updating the epic spec to match so code and spec agree. Minor follow-ups still open: `runTask` skips schema validation when the agent returns no `output`; `auto` throws on a non-ZodObject schema; `renderWorkflow` hand-copies `renderAndSubmit` and drops `trigger`; and there's no usage docs page yet.

`pnpm typecheck`, `check-docs`, `check-llms`, `check-deps` green; testing suite 7/7. Fable re-review to follow (fable re-reviews briefly rate-limited).

roninjin10:
## Final re-review (automated, fable) — head `4c37706`

**Verdict: NITS-ONLY — both MAJOR fixes verified correct, no blockers.**

### Verified

- **renderPrompt re-export is the production renderer.** `packages/testing/src/renderPrompt.ts:6` re-exports `renderPromptToText` from the `@smithers-orchestrator/components` barrel; `packages/components/src/index.js:62` exports it from `components/Task.js:57`, which is the exact function the production `<Task>` path calls (Task.js:284). Verified at runtime: `renderPrompt === renderPromptToText` is `true` — divergence is now structurally impossible. The hand-added `.d.ts` line (`index.d.ts:3306`, `(prompt: unknown): string`) matches the JSDoc (`@param {unknown}` / `@returns {string}`), so a `tsup --dts-only` regen will not drift.
- **fakeAgent safeParse-first disambiguation** (`fakeAgent.ts:97-116`): probed the ambiguous cases empirically.
  - Bare output whose schema has `output`/`text` field names (the original MAJOR): now correctly returned as bare output — `{output:{output:"hello",text:"world"}}`. Fixed.
  - Normal wrapper `{output, text}` against a strict schema: still parsed as wrapper, `text` preserved. Correct (the wrapper's extra keys make the bare parse fail, so precedence never misfires here).
- **fuzzy-select unchanged vs main**: `git diff origin/main...4c3770662 -- '*fuzzy*'` is empty; `apps/cli/src/fuzzy-select.js` and its test are byte-identical to `origin/main`.
- **Tests & gates** (clean worktree at `4c37706`): `bun test packages/testing/tests` → 7 pass / 0 fail; `pnpm -C packages/testing typecheck` clean; `pnpm -C packages/components test` → 327 pass / 0 fail (barrel change safe); root `pnpm typecheck` clean (exercises the new tsconfig paths and the edited `index.d.ts` across all consumers).
- Docs bundles regenerated consistently (export tables + import examples in `llms-full`, `llms-core`, `llms-full-v0.28.0`, cli/skills copies, `package-configuration.mdx`).

### MINOR

1. **Residual false positive in the safeParse-first heuristic** (`packages/testing/src/fakeAgent.ts:105`): when the output schema is *fully permissive or all-optional*, an intended wrapper validates as bare output and is silently swallowed. Repro: schema `z.object({ summary: z.string().optional() })` with script `{ output: { summary: "hi" }, text: "log" }` → returns `{output:{}}` (Zod strips unknown keys) — the intended output, `text`, and `files` are all silently dropped (declared `files` are never written). A strictly-dominating refinement: when the value `hasResponseKeys`, first try `schema.safeParse(value.output)` and prefer the wrapper if that succeeds, else fall back to bare — this keeps the fixed MAJOR case correct (there `value.output` is `"hello"`, which fails the object schema) while fixing the all-optional/loose-schema case. Not blocking: structured task outputs with zero required fields are unusual, and the precedence is documented in the code comment.
2. **`runTask` skips schema validation for non-`output` agent results** (`packages/testing/src/runTask.ts:34-37`): if the agent returns a result without an `output` key (e.g. a text-only fake), the raw envelope (`{text: ...}`) is returned as the task output with no `validateOutput` pass — inconsistent with every other exit path of the function.

### Nits

- Root `package.json` maps `./testing` → `./packages/testing/src/index.ts` directly, while root `tsconfig.json:37` maps `smithers-orchestrator/testing` → `./packages/smithers/src/testing.ts` and every sibling subpath routes through `packages/smithers/src/`. Functionally equivalent (testing.ts is a pure re-export), but one of the two should match the sibling convention.
- `fakeAgent.ts:112`: a non-string `text` in a wrapper is silently dropped rather than rejected.

Scope note: raw-`.ts` shipping / Node-runtime compat is a known, explicitly deferred follow-up per the maintainer — not re-litigated here.

*Automated final re-review by fable (Claude); not a human reviewer.*

PR #520, opened 2026-07-04
fix(cli): stop MCP waits on caller abort

cookesan:
## Summary
- forward MCP request context to semantic tool handlers so caller abort signals reach long waits
- stop watch_run polling, run_workflow startup waits, and ask_human waits when the caller aborts
- normalize raw delete bytes in the CLI fuzzy picker so backspace stays reliable in streamed TTY input

## Tests
- bun test apps/cli/tests/ask-human-mcp.test.js apps/cli/tests/semantic-mcp.test.js apps/cli/tests/semantic-tools-unit.test.js apps/cli/tests/fuzzy-select.test.js
- pnpm -C apps/cli typecheck
- pnpm -C apps/cli test
- pnpm typecheck
- pnpm typecheck:examples
- pnpm typecheck:evals
- pnpm lint
- pnpm check:deps
- pnpm check:docs
- pnpm check:llms
- pnpm check:effect

### DIFF
--- a/apps/cli/src/fuzzy-select.js
+++ b/apps/cli/src/fuzzy-select.js
@@ -144,6 +144,24 @@ function terminalRows() {
 }
 
 /**
+ * Some readline paths can surface delete keys as raw control bytes in the
+ * tracked input. Treat them as edits before running the fuzzy filter.
+ *
+ * @param {string | undefined} input
+ */
+function normalizeTrackedInput(input) {
+    const chars = [];
+    for (const char of String(input ?? "")) {
+        if (char === "\x7f" || char === "\b") {
+            chars.pop();
+        } else {
+            chars.push(char);
+        }
+    }
+    return chars.join("");
+}
+
+/**
  * A type-to-filter picker built on clack's `Prompt` base class. We subclass
  * `Prompt` directly (NOT `SelectPrompt`) with `trackValue=true` so the base
  * pipes readline and keeps `this.value` in sync with the typed query — letters
@@ -199,10 +217,10 @@ class FuzzySelectPrompt extends Prompt {
             if (i >= 0) this.optionCursor = i;
         }
 
-        // Re-filter on every keystroke. Typing AND backspace both edit rl.line,
-        // so the base re-emits "userInput" for both — this covers backspace free.
-        this.on("userInput", () => {
-            this.query = this.userInput ?? "";
+        // Re-filter on every keystroke. Delete keys may arrive either as an
+        // edited readline buffer or as raw control bytes, depending on the stream.
+        this.on("userInput", (input) => {
+            this.query = normalizeTrackedInput(input);
             this.filtered = fuzzyFilter(this.query, this.allOptions);
             if (this.optionCursor > this.filtered.length - 1) {
                 this.optionCursor = Math.max(0, this.filtered.length - 1);
--- a/apps/cli/src/fuzzy-select.js
+++ b/apps/cli/src/fuzzy-select.js
@@ -144,24 +144,6 @@ function terminalRows() {
 }
 
 /**
- * Some readline paths can surface delete keys as raw control bytes in the
- * tracked input. Treat them as edits before running the fuzzy filter.
- *
- * @param {string | undefined} input
- */
-function normalizeTrackedInput(input) {
-    const chars = [];
-    for (const char of String(input ?? "")) {
-        if (char === "\x7f" || char === "\b") {
-            chars.pop();
-        } else {
-            chars.push(char);
-        }
-    }
-    return chars.join("");
-}
-
-/**
  * A type-to-filter picker built on clack's `Prompt` base class. We subclass
  * `Prompt` directly (NOT `SelectPrompt`) with `trackValue=true` so the base
  * pipes readline and keeps `this.value` in sync with the typed query — letters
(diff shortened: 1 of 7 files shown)


roninjin10 (commented):

The abort plumbing (threading `extra.signal` from the MCP SDK into the semantic tools) is the right direction, but as written it violated smithers' core durability contract, so I rebased onto `main` and reworked it.

### Fixed (MAJOR): a caller abort must never cancel a durable run
`waitForTerminal` forwarded `extra.signal` straight into `runWorkflow(..., { signal })`. That means an MCP client crash, timeout, or disconnect would **cancel the whole durable run** — the exact opposite of what smithers promises. A run must outlive the tool call that launched it.

The fix: never forward the caller signal into `runWorkflow`. Instead race the terminal wait against the abort (`settleOrAbort`). If the caller aborts first, we **detach** and report the run as launched-in-background (`launchMode: "background"`); the run keeps executing durably. This matches the tool's own framing ("stop the MCP wait", not "stop the run"). Added a deterministic test: a pre-aborted signal + `waitForTerminal` returns a background summary and the run still finishes (previously it came back `waited`/`cancelled`).

### Fixed (MINOR): orphaned ask_human rows
On abort, `ask_human` left its pending request row behind — a human could later answer a prompt whose asker was long gone. It now calls `adapter.cancelHumanRequest` (which only flips still-`pending` rows, preserving a concurrent real answer), and still returns the structured `aborted`/`blocked` outcome so the caller keeps its `requestId`. Test asserts the row ends `cancelled` and no longer pending.

### Fixed (minors)
- Background observe-wait (`waitForObservedRun`) now returns `null` on abort (reported as launched) instead of throwing.
- Unified the abort error behind `mcpAbortError()`.
- Dropped the dead already-aborted branch in `sleep()` (the top `throwIfAborted` made the trailing guard unreachable); `sleep` now rejects consistently via a single synchronous check.

### Dropped
- The bundled fuzzy-select.js hunk (no real bug under locked `@clack/core@1.4.2`; reverted to `main`).

`pnpm typecheck` green; `ask-human-mcp`, `semantic-tools-unit`, `semantic-mcp` suites pass (19/19). Fable subagent re-review next; merge once clean.

roninjin10:
**Final re-review verdict: NITS-ONLY — the durability contract holds; safe to merge.**

No BLOCKING or MAJOR findings.

**MINOR / nits**

1. `apps/cli/src/mcp/semantic-tools.js:594` — `settleOrAbort`'s pre-aborted fast path (`if (signal.aborted) return Promise.resolve({ aborted: true })`) returns without attaching a rejection handler to `promise`. The only current caller (run_workflow's detach branch) immediately does `void launchPromise.catch(...)`, so nothing leaks today, but a future caller of this generic helper could get an unhandled rejection. Consider adding `promise.then(undefined, () => {})` on that path too, or a JSDoc note that the caller owns the rejection after an aborted result.
2. run_workflow detach response can report `launchMode: "background"` with a terminal `status` (e.g. `"completed"`) and `result: null` when the run finishes before the abort is observed (the pre-aborted-signal case). Accurate and harmless — the caller is gone anyway — just noting the shape.
3. `waitForHumanAnswer` checks abort at the loop top before re-reading the row, so an answer landing in the same window as the abort surfaces as `"aborted"` to the (departed) caller. The pending-only `cancelHumanRequest` guard means the answered row is untouched, so no data is lost. By design per the inline comment.

**What I verified**

- **Durability contract:** `extra?.signal` is never forwarded into `Effect.runPromise(runWorkflow(...))` — grepped every `extra?.signal` use at head; the only touchpoints are the `settleOrAbort` race, `waitForObservedRun`, watch_run's `throwIfAborted`/abortable `sleep`, and `waitForHumanAnswer({ signal })`. All wait-side only.
- **Detached rejection handling:** `settleOrAbort`'s `onAbort` attaches a noop rejection handler before resolving `{ aborted: true }`; the detach branch then logs failures via `void launchPromise.catch(...)`; the non-wait background path retains its pre-existing catch. No unhandled-rejection path in run_workflow.
- **ask_human cancel-only-pending:** `packages/db/src/adapter.js:2017` — `cancelHumanRequest` uses `updateWhere ... request_id = ? AND status = 'pending'`, so a concurrent human answer wins over the cancel. `throwIfAborted` before `insertHumanRequest` avoids creating rows for already-dead callers. The `"aborted"` status is in `askHumanDataSchema`'s enum, so the structured blocked outcome validates.
- **sleep() leak-free:** timer-fire path removes the abort listener; abort path clears the timer; listener registered `{ once: true }`. Engine-side `defaultPollSleep` is unchanged vs main and equally clean.
- **waitForObservedRun:** returns `null` on abort (pre-check + `sleep` catch), so an aborted background launch still reports as launched rather than throwing. Never cancels the run.
- **Tests:** `bun test apps/cli/tests/ask-human-mcp.test.js apps/cli/tests/semantic-tools-unit.test.js apps/cli/tests/semantic-mcp.test.js` at head `366809d74` — **19 pass, 0 fail** (166 expects). The deterministic pre-aborted-signal detach test is real: logs show `runId=semantic-quick-detached ... workflow run finished` — the run completes in the background after the MCP call detaches, and the assertion pins `launchMode === "background"` / `status !== "cancelled"`. The ask_human test asserts the pending row flips to `cancelled` and none remain pending. The watch_run abort test asserts `TASK_ABORTED` and `withDb` cleanup (`cleanupCalls === 1`). The MCP-surface test proves the SDK actually delivers `extra.signal` end to end over a real linked transport.
- **Typecheck:** root `pnpm typecheck` clean at head.
- **`apps/cli/src/fuzzy-select.js`:** zero diff vs `origin/main` in both directions — unchanged, as claimed.

—
*Automated fable re-review (Claude, final pass) — no code was modified.*

PR #521, opened 2026-07-04
fix: surface overdue timers and stabilize fuzzy select

cookesan:
## Summary
- mark waiting timers with past wake times as unhealthy while preserving the blocked timer reason
- add the timer-overdue RunStateView unhealthy variant to DB and DevTools public types
- update run-state docs, generated docs, and the docs checker
- keep fuzzy-select backspace handling from getting stuck when fake keypress streams append DEL
- make compute callback timeouts settle even when the callback ignores cancellation

## Tests
- bun test packages/db/tests/runState-deriveRunState.test.js
- bun test --timeout=60000 --max-concurrency=1 packages/engine/tests/workflow-bridge-compute.test.js
- pnpm -C packages/db test
- pnpm -C packages/devtools test
- pnpm -C apps/cli test
- pnpm -C packages/engine test
- pnpm -C packages/db typecheck
- pnpm -C packages/devtools typecheck
- pnpm -C apps/cli typecheck
- pnpm -C packages/engine typecheck
- pnpm typecheck
- pnpm typecheck:examples
- pnpm typecheck:evals
- pnpm lint
- pnpm check:docs
- pnpm check:llms
- pnpm check:deps
- pnpm check:effect
- git diff --check
- pnpm test

### DIFF
--- a/apps/cli/docs/llms-full.txt
+++ b/apps/cli/docs/llms-full.txt
@@ -9352,6 +9352,7 @@ type ReasonBlocked =
 
 type ReasonUnhealthy =
   | { kind: "engine-heartbeat-stale"; lastHeartbeatAt: string }
+  | { kind: "timer-overdue"; wakeAt: string; overdueMs: number }
   | { kind: "ui-heartbeat-stale";     lastSeenAt: string }
   | { kind: "db-lock" }
   | { kind: "sandbox-unreachable" }
@@ -9360,9 +9361,9 @@ type ReasonUnhealthy =
 
 Timestamps are ISO-8601 strings. Current `computeRunState` / `deriveRunState`
 emits `approval`, `event`, `timer`, `approval-decided-resume-required`, and
-`external-trigger` blocked reasons, plus `engine-heartbeat-stale` unhealthy
-reasons. The other variants are reserved by the public type and may appear on
-future run-state surfaces.
+`external-trigger` blocked reasons, plus `engine-heartbeat-stale` and
+`timer-overdue` unhealthy reasons. The other variants are reserved by the
+public type and may appear on future run-state surfaces.
 
 ## RunStateView
 
@@ -9380,10 +9381,10 @@ type RunStateView = {
 matching pending approval/timer/event context, when a parked `waiting-event`
 run has no event-waiting node, or when that context is supplied to
 `deriveRunState`. A `waiting-*` state can be returned without `blocked` when
-the supporting row is unavailable. `unhealthy` is present for current `stale`
-and `orphaned` results. `recovering` is reserved by the type but is not
-currently emitted by the DB derivation. Terminal states (`succeeded`, `failed`,
-`cancelled`) carry neither.
+the supporting row is unavailable. `unhealthy` is present for current `stale`,
+`orphaned`, and overdue timer results. `recovering` is reserved by the type but
+is not currently emitted by the DB derivation. Terminal states (`succeeded`,
+`failed`, `cancelled`) carry neither.
 
 ## computeRunState
 
@@ -9393,7 +9394,7 @@ import { computeRunState } from "@smithers-orchestrator/db/runState";
 const view = await computeRunState(adapter, runId);
 view.state;       // "running" | ...
 view.blocked;     // present for waiting-* only when backing context is found
-view.unhealthy;   // present for stale/orphaned heartbeat expiry
+view.unhealthy;   // present for stale/orphaned heartbeat expiry or overdue timers
 ```
 
 `computeRunState` is pure over the DB plus the heartbeat / lease signals
--- a/apps/cli/src/fuzzy-select.js
+++ b/apps/cli/src/fuzzy-select.js
@@ -236,20 +236,6 @@ class FuzzySelectPrompt extends Prompt {
             }
         });
     }
-
-    onKeypress(seq, key) {
-        if (key?.name === "backspace" && this.rl) {
-            const delIndex = this.rl.line.indexOf("\x7f");
-            if (delIndex >= 0) {
-                // Fake keypress emitters can append DEL instead of letting readline erase.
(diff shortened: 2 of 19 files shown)


roninjin10 (commented):

The `timer-overdue` unhealthy state is a real, useful signal (a waiting-timer run whose wake time has passed and isn't firing). I rebased onto `main` (the branch was 32 commits behind), added the missing grace threshold, and dropped two things that don't belong.

### Added (MAJOR): grace threshold
As written, the moment `now` passed `firesAtMs` the run was flagged `timer-overdue` — so every run would flicker unhealthy for an instant at its own deadline from ordinary poller lag. Added `RUN_STATE_TIMER_OVERDUE_GRACE_MS` (30s, mirroring the existing `RUN_STATE_HEARTBEAT_STALE_MS` precedent), overridable via a new `timerOverdueGraceMs` input. Your test fixture already used a 30.5s overdue value, which suggested exactly this ~30s intuition — it now passes for the right reason. Added within-grace (20s → healthy), exact-boundary (overdue 0 → healthy), and configurable-grace (0 → flags immediately) tests.

### Dropped (MAJOR): the engine compute-timeout rewrite
I reverted `compute-task-bridge.js` to `main`. The claimed hang does not reproduce: your own strengthened test (compute callback that never resolves, `await new Promise(() => {})`) passes against unmodified `main` — the heartbeat/timeout path already fails the task and moves on. The rewrite added abort-classification and settle-order changes to a hot path for no demonstrated benefit. I kept the strengthened test (it's a genuinely better assertion).

### Fixed (BLOCKER): stale-base docs
The branch regenerated the llms bundles against v0.26.1 — retro-editing a frozen release snapshot and turning `main` (0.28.0) red on `check-llms`. I reverted `docs/llms-full-v0.26.1.txt` to `main` and regenerated the current bundles with `pnpm docs:llms`; the only delta is your `run-state.mdx` `timer-overdue` addition. `check-docs`/`check-llms` both green.

### Fixed (nit): brittle check-docs needle
The grace threshold made `timerRunState(...)` a 4-arg call; the exact-string needle `"return timerRunState(base, pendingTimer, now);"` would have broken. Loosened it to not pin the argument list.

### Follow-up (MINOR, not done here)
The monitor TUI (`monitor.tsx`) gates its unhealthy banner on `stale`/`orphaned`/`recovering` and won't surface `timer-overdue` yet — worth a small follow-up so the new state is actually visible.

Also dropped the fuzzy-select hunk. Fable re-review + merge to follow (fable re-reviews briefly rate-limited).

roninjin10:
**Verdict: NITS-ONLY — the grace-thresholded `timer-overdue` state is correct, the engine bridge rewrite is fully dropped, and all gates pass at head 4eb7bef9b; nothing blocking.**

## What I verified (independent final re-review at 4eb7bef9b)

- **Grace-threshold correctness** — `packages/db/src/runState/deriveRunState.js:107-133`: `overdueMs = now - firesAtMs`; healthy when `overdueMs <= graceMs`, unhealthy only strictly past the grace window. Semantics at the boundaries: `overdue == 0` healthy, `overdue == grace` healthy, `overdue > grace` unhealthy — matches the inline comment ("only unhealthy once overdue by MORE than the grace window") and mirrors the heartbeat-stale pattern. Negative overdue (future timer) trivially healthy. The `blocked: {kind:"timer"}` reason is preserved alongside `unhealthy`, and default `RUN_STATE_TIMER_OVERDUE_GRACE_MS` = 30_000 (`packages/db/src/runState/RUN_STATE_TIMER_OVERDUE_GRACE_MS.js:6`) with per-call override via `timerOverdueGraceMs` (`packages/db/src/runState/DeriveRunStateInput.ts:14`).
- **Tests** — ran `bun test packages/db/tests/runState-deriveRunState.test.js` at the PR head: 23 pass, 0 fail. New cases cover overdue-past-grace (30.5s → unhealthy with exact `overdueMs: 30500`), within-grace (20s → healthy), exactly-at-wake (`overdue == 0` → healthy), configurable grace (`graceMs: 0`, 1ms overdue → unhealthy), and the existing future-timer test now asserts `unhealthy` is undefined.
- **Engine bridge untouched** — `git diff origin/main...4eb7bef9b -- packages/engine/src/effect/compute-task-bridge.js` is empty; the only engine change is the strengthened test (`packages/engine/tests/workflow-bridge-compute.test.js:180-203`), which replaces `Bun.sleep(100)` with a never-resolving promise plus a `callbackStarted` assertion. Ran it at head: 4 pass, 0 fail — confirms main's bridge already handles the timeout and the dropped rewrite was unnecessary.
- **Gates** — `node scripts/check-docs.mjs` and `node scripts/check-llms.mjs` both pass at head (in a full-deps worktree), including the run-state derivation-contract check with the de-brittled needles (`scripts/check-docs.mjs:904-925` — the new `timerRunState(...)` needle plus a forbidden needle guarding the stale doc line). `tsc --noEmit -p packages/db` clean.
- **Type surface consistency** — `timer-overdue` added identically in `packages/db/src/runState/ReasonUnhealthy.ts:3`, `packages/devtools/src/DevToolsSnapshot.ts:31`, `packages/devtools/src/index.d.ts:214-217`, and `docs/runtime/run-state.mdx`. No stale `0.26.1` strings in the regenerated llms bundles (regen is at 0.28.0). Grepped all UI/server packages: no consumer switches on `unhealthy.kind`, so the new variant cannot break existing rendering.

## Nits (non-blocking)

1. **Exact-boundary case untested** — tests cover 0, within-grace (20s), and just-over (30.5s), but not `overdueMs === graceMs` exactly (healthy per `<=` at `deriveRunState.js:122`). A one-liner would pin the documented "MORE than" semantics.
2. **monitor.tsx does not render `timer-overdue`** — already acknowledged as an acceptable follow-up; confirmed no UI package consumes `unhealthy` today, so this is invisible until a renderer lands.
3. **`Math.floor(overdueMs)`** at `deriveRunState.js:130` is a no-op for integer ms inputs — harmless defensive rounding, fine to keep.

---
*Automated re-review by fable (Claude Fable 5), final independent pass.*

PR #641, opened 2026-07-07
🐛 fix(server): declare @smithers-orchestrator/graph as a runtime dependency

JasonVranek:
## Summary

`packages/server/src/index.js` imports `escapeSmithersDir` from `@smithers-orchestrator/graph` at runtime, but the package is declared only under `devDependencies` — so the root `pnpm test` gate fails on `main` at `check-dependency-boundaries`:

```
Dependency boundary check failed: undeclared imports found.
- packages/server/src/index.js imports @smithers-orchestrator/graph/escapeSmithersDir; declare @smithers-orchestrator/graph as a workspace dependency in dependencies.
```

Moved the entry to `dependencies` (`workspace:*`, alphabetical slot) and regenerated the lockfile. `check-dependency-boundaries` now passes for all 65 packages.

(Found while pre-flighting a larger memory PR; splitting this out so it's a 30-second review. One more pre-existing red on `main` we did NOT touch: `check-llms` reports the generated `llms-full.txt` bundles out of date — left alone since regeneration looks like your release flow.)


JasonVranek:
Heads-up on the red CI here — it's a second layer of pre-existing `main` breakage that this fix *unmasks*, not something this diff causes:

- On `main` (and on #642), every CI test/typecheck job dies at the dependency-boundary gate with exactly the error this PR fixes — so CI never reaches the build step. Verified on the last three `main` commits: same five failing jobs, same gate error.
- With this fix applied, the gate passes (`Dependency boundary check passed for 65 package(s)` in this PR's jobs) and CI advances to the next latent failures on `main`'s tip, which a one-line dependency move can't touch:
  - `packages/gateway-react` dts build: `TS7016: Could not find a declaration file for module '@smithers-orchestrator/gateway/rpc'`
  - `packages/gateway` typecheck: `scripts/generate-openapi.ts(116,10): error TS4058 ... 'JsonSchema$1'`

Happy to leave those to you (they look like a missing `.d.ts`/types export on the gateway `rpc` subpath), or send a follow-up if useful.

JasonVranek:
Pushed `bccf40c` — the root cause of the intermittent `gateway-react` TS7016 (the failure CI hits once this PR's dep fix lets it reach the build step):

**`packages/gateway`'s own dts build races itself.** tsup runs the four per-subpath configs concurrently against the shared `src` outDir, and each config's `clean` races the sibling configs' emits — so `tsup --dts-only` nondeterministically exits 0 with `src/index.d.ts` / `src/rpc/index.d.ts` **missing** (`silent: true` hides it). Reproduced locally at 2-in-8:

```
run 5: exit=0 rpc.d.ts=present idx=MISSING
run 7: exit=0 rpc.d.ts=MISSING idx=MISSING
```

When `rpc/index.d.ts` loses the race, gateway-react's dts build later fails TS7016 on `@smithers-orchestrator/gateway/rpc` — matching the intermittent red on `main` since the run at `1343b0bd4` (and why parallel matrix jobs disagree on the same commit).

**The fix** keeps the stale-declaration pruning the `clean` existed for, but moves it *before* tsup via the existing `scripts/clean-dts.mjs` (the `packages/testing` pattern), and drops the racy in-config `clean`. 8/8 stable after; gateway-react builds green downstream; `check:openapi`, gateway tests, and root typecheck all pass locally (with your `RpcFrameSchema` fix — nice, I dropped my equivalent in its favor).

Also: apologies for a brief force-push that momentarily replaced `0640187` while I was syncing — restored within a minute; this push is a plain fast-forward on top of your head.

PR #642, opened 2026-07-07
✨ feat(memory): durable knowledge — provenance, append-only notes with supersession, status gating, lazy FTS

JasonVranek:
## Summary

`@smithers-orchestrator/memory` today is a namespaced KV fact store (`MemoryFact
{namespace, key, valueJson, schemaSig, ttlMs}`, upsert semantics). This PR adds a small
set of GENERIC properties that let the same package carry durable KNOWLEDGE — append-only,
auditable, human-gateable — while staying policy-free. The design is staged so each piece
is independently reviewable and independently valuable; happy to split this into three
PRs along those lines if you prefer.

Everything here is implemented, tested (23 store-level tests + two engine-level e2e
suites, all keyless/CI-safe), and validated by a production downstream that migrated its
memory layer onto this exact schema before submission.

## Motivation (the general pattern)

Every long-lived agent system grows this shape: the ecosystem is full of memory layers
bolted BESIDE orchestrators precisely because engines offer only KV/scratch state. Teams
accumulate decisions, invariants, and learnings across runs, and the engine owns the two
things such a store needs most and userland cannot fake: the RUN COORDINATE SYSTEM
(runId/nodeId/iteration) and DURABLE EXECUTION for the workflows that maintain memory.
We are one worked existence proof (we built ~15 side tables downstream before concluding
the bottom four properties belonged in the engine) — but the pattern is your user base at
large, not one downstream.

## What's in the PR

**Stage 1 — provenance, explicitly stamped (the keystone).** Nullable
`run_id/node_id/iteration` on memory writes: facts record the last writer (upsert
semantics), messages gain the missing `iteration` alongside their existing
run_id/node_id, and notes (below) carry the full coordinate. The write API takes an
OPTIONAL provenance parameter — explicitly not ambient context (ambient scope dies across
agent boundaries; we have paid for that lesson). Callers that hold the coordinates pass
them: a workflow build function closes over `ctx.runId`; an agent's memory tool closes
over the coordinates at construction, which also means the model cannot fabricate where a
note came from. This mirrors a convention the engine already has — every structured-output
table row is auto-stamped `run_id/node_id/iteration` — extended to memory rather than
invented. Zero change for existing callers (the 4-arg `setFact` call sites run untouched).

**Stage 2 — append-only notes with supersession + status.** A SIBLING record type, not a
versioned mode on facts (considered and rejected: two write disciplines on one table means
a caller who upserts what should have been a note silently destroys history; a sibling
type makes the invariant structural):

```
MemoryNote { id, namespace, body, kind?, tags?, author?, status default 'accepted',
             status_changed_at_ms?, created_at_ms, run_id?, node_id?, iteration? }
_smithers_memory_note_supersessions { note_id, supersedes_id, created_at_ms }
```

- `kind`/`tags`/`author` are optional, policy-free labels the engine never interprets.
- Supersession is a JUNCTION (note_id = the superseder; one accepted synthesis → N rows).
  Considered and rejected: a JSON-array column (JSON in the hot read path) and a
  `superseded_by` column on the superseded row (mutates historical rows, violating the
  append-only invariant this stage exists to establish).
- Notes carry NO ttl — timer-expiring append-only knowledge would silently vanish a
  ratified decision; knowledge dies by supersession or rejection, not by clock.
- `status` is free-form and THE ONE deliberate exception to append-only: the human gate
  writes an answer about an existing note (`setNoteStatus`, stamped with
  `status_changed_at_ms`). Transition history is userland (a ratifications table keyed by
  note id); the engine keeps current status + the latest transition time.
- **The DEFAULT READ CONTRACT, requested as a documented stability contract:** default
  reads return notes that are (a) not superseded by an ACCEPTED note AND (b)
  status=accepted; callers pass filters to widen. Chain semantics are load-bearing for
  human gating: a note is hidden ONLY by an accepted superseder — a pending or rejected
  superseder hides nothing, so propose-supersession-then-reject leaves the original live.
  The whole contract compiles to one NOT-EXISTS predicate over the junction (no recursion);
  chains behave (A<B<C reads [C]; rejecting C revives B while A stays hidden by B).
- `saveNote` is idempotent on id (INSERT-OR-IGNORE — crash-resume/replay safe, mirroring
  `saveMessage`), and writes its supersession edges atomically with the note.

**Stage 3 — text search, lazy and opt-in.** FTS5 over note bodies, opt-in PER NAMESPACE
KIND: `enableNoteSearch(kind)` creates the FTS artifacts on first call and backfills;
until then no FTS table exists at all, so KV-only users pay zero write amplification and
no migration. Searching a never-enabled kind fails loud with a pointer to
`enableNoteSearch` — never silently empty (silent-empty reads poison corpora; paid for
that lesson too). KV answers "what is X"; knowledge needs "what do we know about X".

**Schema/migrations** follow the house pattern exactly: drizzle defs in
`packages/db/src/internal-schema/`, idempotent DDL in `sql-message-storage.js`, and three
ledger migrations (`0021`/`0022` additive columns, `0023` new tables), validated in-place
on a database with pre-existing facts (rows intact through the existing column-migration
machinery). The SQLite→PGlite store migrator copies the new tables row-for-row with no
special-casing.

**The example** — `examples/incident-runbook-memory.jsx` (+ prompts), following the
existing examples conventions: an on-call helper that gets smarter every incident. Recall
(FTS + the default read contract — the responder only ever sees live knowledge; the payoff
of every previous run) → triage grounded in prior lessons → bank what this incident taught
(the `record_lesson` tool closes over the run's provenance) → when lessons pile up,
propose ONE pending runbook rule superseding them → a `needsApproval` gate whose compute
is the single mutable status write. Approving puts the rule live and retires the lessons
atomically; denying changes nothing. Every "memory feature" in it is a plain Task — the
engine additions are storage properties only.

## Validation

- **23 store-level tests** (`packages/memory/tests/notes.test.js`): provenance on every
  write path, chain semantics (pending/rejected superseders hide nothing; rejection
  revives), lazy FTS (zero artifacts until opt-in; per-kind isolation; fail-loud),
  idempotency, and the in-place migration of a pre-change database.
- **Consolidation e2e under the real engine**
  (`packages/smithers/tests/memory-consolidation.e2e.test.jsx`): `runWorkflow` parks at
  the approval gate, `approveNode`/`denyNode` drive both paths, resume completes. Mid-gate
  the pending synthesis hides nothing; approval flips visibility atomically with the one
  status write; denial leaves the corpus untouched.
- **The example's own e2e** (`packages/smithers/tests/incident-runbook-memory.e2e.test.jsx`):
  the example flow verbatim with fake agents, FIVE runs on one db — lessons accumulate
  with per-run provenance, the threshold crosses, the gate parks/approves/resumes, and the
  fifth run's recall serves one rule where three lessons were. Keyless, CI-safe.
- **A production downstream runs on this schema.** Before submitting, we migrated our own
  memory layer (a task/knowledge system built on Smithers) onto these exact tables:
  row-for-row corpus migration, read paths re-expressed over the substrate, full read
  parity (ids and ordering). The migration diff surfaced a real bug in our hand-rolled layer that this
  contract prevents by construction: our supersession hide-predicate was status-blind, so
  a rejected proposal kept its target hidden forever — silent knowledge loss we shipped
  and only found by diffing against these semantics. That is the argument for stating the
  read contract once, in the engine, verified. Since migrating we have run a full overnight
  multi-agent dogfood (~30 workflow runs, task/knowledge writes throughout) on the
  substrate-backed store without an anomaly.

Six higher-level memory roles we run downstream — inheritance, distillation, calibration,
association, narrative, invalidation — all sit on top of these properties as libraries and
ordinary workflows with ZERO additional engine schema: the substrate is complete for a
demanding consumer without policy creep.

## Non-goals (stay userland, deliberately)

Ranking/relevance; embeddings/semantic search (userland vectors keyed by note id);
consolidation/summarization (ordinary workflows); trust/calibration models (userland
tables keyed by note id + provenance); context injection (prompt composition in the build
fn); retention beyond ttlMs.

## Notes for reviewers

- Heads-up, unrelated to this diff: two of the root `pnpm test` meta-checks currently fail
  on `main` itself (the dependency-boundary check — `packages/server` imports
  `@smithers-orchestrator/graph` at runtime but declares it only in devDependencies — and
  stale generated `llms-full.txt` artifacts). #641 is a separate one-line fix for the
  former; the llms artifacts we left to you since regeneration is your release flow.

- The public API surface grew by six methods (`saveNote`, `getNote`, `listNotes`,
  `setNoteStatus`, `enableNoteSearch`, `searchNotes`) on MemoryStore/MemoryService; the
  exact-surface test in `service.test.js` and the migrate-store schema-head assertions
  were updated accordingly.
- Two small follow-ups we'd take with guidance: an optional `createdAtMs` on
  `SaveNoteInput` (importers/migrations need to preserve timestamps; `saveMessage` already
  has it), and a `memoryNoteWrites` metric in the observability package mirroring
  `memoryFactWrites`.
- Postgres: table creation goes through the existing `translateDdl` path; FTS is
  SQLite-only (matching the engine's sqlite posture), and `enableNoteSearch` is where a
  tsvector implementation would slot later.


roninjin10:
I love this idea!

JasonVranek:
Note on this PR's red CI: all four failing jobs stop at the dependency-boundary gate — the pre-existing `main` failure that #641 fixes (`packages/server` imports `@smithers-orchestrator/graph` without declaring it in `dependencies`). `main`'s own CI fails the identical five jobs at the same gate on its last three commits; this diff never touches `packages/server`. With #641 merged (or rebased in), the gate passes — see the analysis on #641 for the two deeper pre-existing failures CI then reaches.

PR #646, opened 2026-07-08
🐛 fix(sandbox): best-effort failure-path bookkeeping so it never masks the primary error

### LINKED ISSUE #524: 🐛 sandbox: failure-path DB writes in executeSandbox can mask the original error
**What happens**
In `packages/sandbox/src/execute.js:707-736` (the `catch` block of `executeSandbox`), `await adapter.upsertSandbox(...)` and `await emitSandboxEvent(... SandboxFailed ...)` run unguarded before `throw error`.

If the bookkeeping write fails — DB closed after a cancel, disk error, connection loss — that secondary exception propagates instead of the primary sandbox error, so the caller sees e.g. a DB-closed error rather than the real failure. Worse, the throw from `upsertSandbox` also skips the `SandboxFailed` event and the failure heartbeat, leaving no durable record of why the sandbox died. The adjacent `finally` block (lines 737-762) already does this correctly: cleanup failures are caught and logged as warnings so they never mask the primary result.

**Expected behavior**
Best-effort failure bookkeeping: wrap the upsert/event/heartbeat in try/catch (log a warning like the finally block does) so the original error always rethrows.

roninjin10:
Closes #524

## Root cause

In `packages/sandbox/src/execute.js`, the `catch` block of `executeSandbox` performed three sequential DB/event writes on the failure path — `adapter.upsertSandbox(...)` (mark sandbox `failed`), `emitSandboxEvent(...)` (emit `SandboxFailed`), and `runtime.heartbeat(...)` (report `stage: "failed"`) — with no error handling around any of them. If the underlying DB was already closed or unreachable (e.g. after a cancel, or a disk/connection failure), any of these bookkeeping writes could throw. Because they ran unguarded inside the `catch`, that secondary failure would propagate instead of the *original* sandbox error, hiding the real cause of the failure from callers.

## Fix

Wrapped each of the three bookkeeping calls in its own `try/catch` so a secondary failure in one write can never suppress another, and — critically — can never mask or replace the primary error being handled. Each bookkeeping failure is logged as a warning (`logWarning("sandbox failure bookkeeping failed", ...)`) so it's still observable, but the original `error` is always the one re-thrown via the existing `throw error;` at the end of the catch block. This mirrors the same best-effort rationale already used in the adjacent `finally` cleanup block.

Added a regression test in `packages/sandbox/tests/execute.test.js` that closes the DB mid-run before the primary provider error is thrown, then asserts the primary error (`"primary sandbox failure"`) is what surfaces from `executeSandbox`, and that the resulting bookkeeping failure is logged rather than swallowed.

## Note

This PR will be **restacked** onto a PR train — its base branch may change as part of that process.

### DIFF
--- a/packages/sandbox/src/execute.js
+++ b/packages/sandbox/src/execute.js
@@ -732,33 +732,67 @@ export async function executeSandbox(options) {
         });
     }
     catch (error) {
-        await adapter.upsertSandbox({
-            runId: runtime.runId,
-            sandboxId: options.sandboxId,
-            runtime: selectedRuntime,
-            remoteRunId: null,
-            workspaceId: handle?.workspaceId ?? null,
-            containerId: handle?.containerId ?? null,
-            configJson,
-            status: "failed",
-            shippedAtMs: createdAtMs,
-            completedAtMs: nowMs(),
-            bundlePath: handle?.resultPath ?? null,
-        });
-        await emitSandboxEvent(runtimeDb, {
-            type: "SandboxFailed",
-            runId: runtime.runId,
-            sandboxId: options.sandboxId,
-            runtime: selectedRuntime,
-            error: errorToJson(error),
-            timestampMs: nowMs(),
-        });
-        runtime.heartbeat({
-            sandboxId: options.sandboxId,
-            stage: "failed",
-            progress: 100,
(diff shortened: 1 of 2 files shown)


roninjin10:

PR #647, opened 2026-07-08
🐛 fix(sandbox): preserve real shippedAtMs through completion/failure upserts

### LINKED ISSUE #525: 🐛 sandbox: completion/failure upserts clobber shippedAtMs with the creation timestamp
**What happens**
`packages/sandbox/src/execute.js` records the real ship time on the "shipped" upserts (`shippedAtMs: nowMs()` at lines 547 and 635), but the shared completion tail overwrites it: `finalizeSandboxBundle` writes `shippedAtMs: params.createdAtMs` (line 405), and the failure upsert in the catch block writes `shippedAtMs: createdAtMs` (line 717).

**Why it's wrong**
Since upsertSandbox rewrites the whole row, every completed or failed sandbox ends up with `shippedAtMs === createdAtMs`, erasing the actual ship timestamp. Any consumer computing ship latency (created→shipped) or execution duration (shipped→completed) from the sandbox row gets fabricated values; the correct timestamp only survives for sandboxes that die between ship and completion bookkeeping.

**Expected behavior**
Preserve the real shipped time on completion/failure — thread the shipped timestamp through to `finalizeSandboxBundle` and the catch-block upsert (or make the upsert not overwrite an existing non-null `shippedAtMs`).

roninjin10:
Closes #525

## Root cause

`executeSandbox` in `packages/sandbox/src/execute.js` upserts the sandbox row multiple times as a run progresses through shipped → completed/failed. The `shippedAtMs` column is meant to record the instant the sandbox was actually shipped, but two of the later upserts were writing the wrong value instead of that captured instant:

- The completion path (`finalizeSandboxBundle`, and its caller) was passing `createdAtMs` (the sandbox's creation time) as `shippedAtMs`.
- The failure path's upsert was also passing `createdAtMs` for `shippedAtMs`.

Since the "shipped" upsert had already written a freshly-computed `nowMs()` for `shippedAtMs` in the shipped stage, subsequent completion/failure upserts clobbered that real value with an unrelated timestamp (`createdAtMs`), losing the true "when did this sandbox actually ship" signal.

## Fix

Introduced a `shippedAtMs` local variable in `executeSandbox`, set it once (to `nowMs()`) at the moment the "shipped" stage progress event fires and the shipped-status upsert runs, and threaded that captured value through to `finalizeSandboxBundle` and the failure-path upsert instead of re-deriving or reusing `createdAtMs`. This preserves the real shipped timestamp through every later completion/failure upsert. Covered by updated assertions in `packages/sandbox/tests/execute.test.js`.

## Strategy

opus-sandwich

Note: this PR will be RESTACKED onto a PR train — its base branch may change after this is opened.

### DIFF
--- a/packages/sandbox/src/execute.js
+++ b/packages/sandbox/src/execute.js
@@ -351,7 +351,7 @@ function isSandboxActive(status) {
  *   options: ExecuteSandboxOptions;
  *   selectedRuntime: string;
  *   configJson: string;
- *   createdAtMs: number;
+ *   shippedAtMs: number | null;
  *   childStartedMs: number;
  *   validated: import("./ValidatedSandboxBundle.ts").ValidatedSandboxBundle;
  *   remoteRunId: string | undefined;
@@ -423,7 +423,7 @@ async function finalizeSandboxBundle(params) {
         containerId: params.containerId,
         configJson,
         status: validated.manifest.status,
-        shippedAtMs: params.createdAtMs,
+        shippedAtMs: params.shippedAtMs,
         completedAtMs: nowMs(),
         bundlePath: validated.bundlePath,
     });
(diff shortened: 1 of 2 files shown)


roninjin10:

PR #648, opened 2026-07-08
🐛 fix(agent-eliza): skip unstatable dir entries in loadWorkflowsFromDir

### LINKED ISSUE #530: 🐛 agent-eliza: one dangling symlink in a workflows dir aborts the entire loadWorkflowsFromDir
**What happens**
`packages/agent-eliza/src/conventions/loader.js:169`: `loadWorkflowsFromDir` iterates `readdirSync(dir)` and calls `statSync(filePath)` unguarded. A dangling symlink — or a file deleted between readdir and stat — makes `statSync` throw ENOENT, rejecting the whole call and losing every valid workflow in the directory.

**Why it's wrong**
Every other per-file failure in this loader is handled with a diagnostic and `continue` (unimportable file at lines 180-187, unrecognizable export at lines 190-197). One broken symlink in `~/.smithers/workflows` should not take down workflow discovery for the whole directory (and via `loadWorkflows`, potentially mask higher/lower-precedence sources with a rejection).

**Expected behavior**
Wrap the stat in try/catch: skip the entry (or push a per-file `error` diagnostic) and continue, matching the loader's established pattern.

roninjin10:
Closes #530

## Root cause

`loadWorkflowsFromDir` called `statSync(filePath)` on every directory
entry and immediately threw if it failed. A dangling symlink (a
symlink whose target doesn't exist) makes `statSync` throw `ENOENT`,
so a single broken symlink in a workflow directory aborted loading the
*entire* directory — every sibling workflow file failed to load too,
even though only one entry was actually broken.

## Approach

In `packages/agent-eliza/src/conventions/loader.js`, wrap the
`statSync` call in a try/catch. On failure, push an `error` diagnostic
naming the unstatable path and `continue` to the next entry instead of
letting the exception propagate. Valid sibling workflows in the same
directory now still load, and the bad entry surfaces as a diagnostic
instead of a hard crash.

Added `packages/agent-eliza/tests/loader-coverage.test.ts` coverage:
creates a dangling symlink alongside a valid workflow file and asserts
the valid workflow still loads while the symlink produces exactly one
error diagnostic.

## Strategy

Implemented using the **opus-sandwich** strategy.

## Note

This PR will be **restacked** onto a PR train — its base branch may
change as other issue PRs land ahead of it.

### DIFF
--- a/packages/agent-eliza/src/conventions/loader.js
+++ b/packages/agent-eliza/src/conventions/loader.js
@@ -166,7 +166,18 @@ export async function loadWorkflowsFromDir({ dir, source = "unknown" }) {
   const entries = readdirSync(dir);
   for (const entry of entries) {
     const filePath = join(dir, entry);
-    if (!statSync(filePath).isFile()) continue;
+    let stats;
+    try {
+      stats = statSync(filePath);
+    } catch (err) {
+      diagnostics.push({
+        type: "error",
+        message: `Could not stat directory entry (${source}): ${filePath} (${err instanceof Error ? err.message : String(err)})`,
+        path: filePath,
+      });
+      continue;
+    }
+    if (!stats.isFile()) continue;
     if (!WORKFLOW_EXTENSIONS.has(extname(entry))) continue;
 
     // Read source text and parse frontmatter BEFORE dynamic import. The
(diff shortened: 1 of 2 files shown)


roninjin10:

PR #649, opened 2026-07-08
🐛 fix(agent-eliza): guard scalar frontmatter tags/aliases in loader

### LINKED ISSUE #531: 🐛 agent-eliza: unvalidated frontmatter tags/aliases crash the workflow formatter (`.join` on a string)
**What happens**
`packages/agent-eliza/src/conventions/loader.js:125-126`: `buildDefinition` typeof-guards `name`/`description`/`system`/`version` from frontmatter but casts `frontmatter.tags` and `frontmatter.aliases` straight to `string[] | undefined` with no `Array.isArray` check.

YAML frontmatter like `tags: foo` parses to a plain string. It flows into `WorkflowDefinition.tags`, passes the formatter's guard (`w.tags && w.tags.length > 0` is true for a non-empty string), and then `w.tags.join(", ")` throws `TypeError: w.tags.join is not a function` (`packages/agent-eliza/src/conventions/formatter.js:49-54`) — one malformed user workflow file crashes prompt formatting for the whole listing.

**Expected behavior**
`Array.isArray(frontmatter.tags) ? frontmatter.tags.filter(t => typeof t === "string") : undefined` (same for aliases), mirroring the validation applied to the other frontmatter fields.

roninjin10:
Closes #531

## Root cause

In `packages/agent-eliza/src/conventions/loader.js`, `buildDefinition()` read
`frontmatter.tags` / `frontmatter.aliases` straight out of the parsed YAML
frontmatter and cast them to `string[]` without validating the actual runtime
shape. If a workflow's frontmatter declared `tags` or `aliases` as a scalar
(e.g. `tags: foo` instead of `tags: [foo]`), the loader would happily pass the
raw string through as the "array", corrupting downstream consumers (e.g.
`formatWorkflowsForPrompt` in `packages/agent-eliza/src/conventions/formatter.js`)
that assume an actual array and call array methods on it.

## Approach

Guard both fields at the point they're read off frontmatter:
- Only accept the frontmatter value when it's actually `Array.isArray(...)`.
- Filter the array to string entries only, dropping any non-string members
  (e.g. `tags: [1, "ok"]` → `["ok"]`).
- Anything else (a bare scalar, an object, etc.) coerces to `undefined`,
  matching the existing `raw.tags`/`raw.aliases` fallback semantics already
  used elsewhere in `buildDefinition()`.

Added regression coverage in `packages/agent-eliza/tests/loader-coverage.test.ts`
covering both the scalar-frontmatter case and the mixed-type-array case.

## Strategy

Implemented via the "opus-sandwich" strategy.

## Note

This PR will be **restacked** onto a PR train — its base branch may change
before merge.

### DIFF
--- a/packages/agent-eliza/src/conventions/loader.js
+++ b/packages/agent-eliza/src/conventions/loader.js
@@ -122,8 +122,16 @@ function buildDefinition(filePath, baseDir, exported, source, frontmatter) {
       raw.description ??
       (typeof frontmatter.description === "string" ? frontmatter.description : undefined) ??
       "",
-    tags: raw.tags ?? /** @type {string[] | undefined} */ (frontmatter.tags),
-    aliases: raw.aliases ?? /** @type {string[] | undefined} */ (frontmatter.aliases),
+    tags:
+      raw.tags ??
+      (Array.isArray(frontmatter.tags)
+        ? frontmatter.tags.filter((t) => typeof t === "string")
+        : undefined),
+    aliases:
+      raw.aliases ??
+      (Array.isArray(frontmatter.aliases)
+        ? frontmatter.aliases.filter((a) => typeof a === "string")
+        : undefined),
     disableModelInvocation:
       raw.disableModelInvocation ??
       (typeof frontmatter["disable-model-invocation"] === "boolean"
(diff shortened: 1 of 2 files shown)


roninjin10:

PR #650, opened 2026-07-08
🐛 fix(memory): map Summarizer agent.run rejection to SmithersError

### LINKED ISSUE #532: 🐛 memory: Summarizer agent.run failures escape as UnknownException instead of SmithersError
**What happens**
`packages/memory/src/Summarizer.js:73`: `yield* Effect.tryPromise(() => agent.run(prompt))` uses the single-argument form, so a rejected `agent.run` surfaces as `UnknownException` in the Effect error channel.

**Why it's wrong**
Both the local JSDoc (`@returns {Effect.Effect<void, SmithersError>}`) and the `MemoryProcessor` contract (`packages/memory/src/MemoryProcessor.ts:8`: `processEffect: (store) => Effect.Effect<void, SmithersError>`) promise a typed `SmithersError` channel. Callers matching on SmithersError code/details get an untyped `UnknownException` instead. The store layer already does this correctly (`MemoryStoreLive.js` uses `Effect.tryPromise({ try, catch })` with error mapping). The mismatch is invisible to `tsc` today only because `checkJs` is not enabled in packages/memory.

**Expected behavior**
`Effect.tryPromise({ try: () => agent.run(prompt), catch: (cause) => toSmithersError(cause, ...) })` so summarizer agent failures stay inside the declared error type.

roninjin10:
Closes #532

## Root cause

`Summarizer` in `packages/memory/src/Summarizer.js` wrapped the agent call with `Effect.tryPromise(() => agent.run(prompt))`, using the single-function-argument form. That form has no `catch` mapper, so when `agent.run(prompt)` rejects, Effect surfaces it as an untyped `UnknownException` instead of a `SmithersError`. Any caller pattern-matching on `SmithersError` codes (retry/backoff logic, error reporting, etc.) silently fails to recognize agent-run failures during summarization.

## Fix

- `packages/memory/src/Summarizer.js`: switched to the `{ try, catch }` form of `Effect.tryPromise`, routing the rejection through `toSmithersError` (from `@smithers-orchestrator/errors/toSmithersError`) with `code: "AGENT_CLI_ERROR"` and `details: { threadId }` for traceability.
- `packages/memory/tests/processors.test.js`: added a regression test that fails the agent's `run` and asserts the resulting `Effect` exit is a failure whose cause is a typed `SmithersError` with code `AGENT_CLI_ERROR`, not an `UnknownException`.

## Strategy

Built using the **opus-sandwich** strategy.

## Notes

This PR will be **restacked** onto a PR train as part of a batch of related issue fixes in this repo — its base branch may change from `main` to another in-flight branch before merge.

### DIFF
--- a/packages/memory/src/Summarizer.js
+++ b/packages/memory/src/Summarizer.js
@@ -1,4 +1,5 @@
 import { Effect } from "effect";
+import { toSmithersError } from "@smithers-orchestrator/errors/toSmithersError";
 /** @typedef {import("./MemoryProcessor.ts").MemoryProcessor} MemoryProcessor */
 /** @typedef {import("./store/MemoryStore.ts").MemoryStore} MemoryStore */
 /** @typedef {import("@smithers-orchestrator/errors/SmithersError").SmithersError} SmithersError */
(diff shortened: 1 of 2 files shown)


roninjin10:

PR #651, opened 2026-07-08
🐛 fix(accounts): preserve unknown-provider entries across add/remove rewrites

### LINKED ISSUE #529: 🐛 accounts: unknown-provider entries are silently deleted from accounts.json by any unrelated add/remove
**What happens**
`packages/accounts/src/parseAccountsFile.js:86-93` skips entries with unrecognized providers (e.g. a legacy `gemini` subscription account) with a warning, explicitly "rather than failing the whole file". But `addAccount` (`packages/accounts/src/addAccount.js:40-57`) and `removeAccount` (`packages/accounts/src/removeAccount.js:19-25`) do `readAccounts → mutate → writeAccounts` of the entire parsed file — so the next time the user adds or removes ANY unrelated account, the skipped legacy entry is permanently erased from disk.

**Why it's wrong**
The parse-time design is documented as tolerant/non-destructive, but the rewrite path silently destroys data the user may want to migrate (notably the legacy entry's `configDir`, which points at real credentials on disk). The only signal is a once-per-process console.warn about *skipping*, which says nothing about deletion.

**Expected behavior**
Preserve unrecognized raw entries when rewriting accounts.json (carry them through readAccounts as opaque passthrough rows), or at minimum warn loudly that the rewrite is deleting them.

roninjin10:
Closes #529

**Root cause:** `parseAccountsFile.js` correctly skipped account entries with an unrecognized `provider` (e.g. a legacy subscription provider removed from `VALID_PROVIDERS`), logging a warning and excluding them from the parsed `accounts` array. But `addAccount.js` and `removeAccount.js` both round-trip through `writeAccounts({ version: 1, accounts: next }, env)` using only the filtered `accounts` array — so any add or remove operation silently rewrote `accounts.json` and dropped those unknown-provider entries from disk entirely, even though they were never touched by the operation.

**Approach:**
- `AccountsFile.ts`: add an optional `unknownAccounts?: unknown[]` field to the in-memory shape, to carry raw unrecognized entries alongside the parsed/validated `accounts`.
- `parseAccountsFile.js`: collect skipped entries into `unknownAccounts` and return them on the parsed result (only when non-empty).
- `addAccount.js` / `removeAccount.js`: thread `existing.unknownAccounts` through into the `writeAccounts` call so a rewrite of unrelated accounts never loses them.
- `writeAccounts.js`: re-append `contents.unknownAccounts` onto the on-disk `accounts` array at serialization time, so the file always contains every entry it started with, known or not.

This PR will be **restacked onto a PR train** — its base branch may change as part of that process.


roninjin10:

PR #652, opened 2026-07-08
🐛 fix(memory): preserve original createdAtMs on saveMessage re-save

### LINKED ISSUE #533: 🐛 memory: saveMessage re-save overwrites createdAtMs, breaking message ordering on crash-resume/replay
**What happens**
`packages/memory/src/store/MemoryStoreLive.js:229-258`: `saveMessageEffect` computes `createdAtMs = msg.createdAtMs ?? nowMs()` and includes `createdAtMs` in the `onConflictDoUpdate` set.

The comment (lines 233-236) says re-saving the same message id on crash-resume, deterministic replay, or fork/restore "must be a safe no-op upsert". But if the re-save omits `createdAtMs` (the field is optional in the signature), the conflict update stamps a NEW `nowMs()` onto the existing row — a replayed old message jumps to the end of `listMessages` (ordered by `createdAtMs`, line 272), corrupting conversation order and downstream summarization windows (Summarizer slices oldest-first).

**Expected behavior**
Exclude `createdAtMs` from the conflict-update set so the original insert timestamp is preserved — making the upsert actually idempotent as documented.

roninjin10:
Closes #533

## Root cause

`saveMessage` in `packages/memory/src/store/MemoryStoreLive.js` upserts into
`smithersMemoryMessages` via `onConflictDoUpdate`. The conflict-update `set`
clause included `createdAtMs`, so re-saving an existing message id (e.g. a
crash-resume / replay caller that only has the id + updated content, without
re-supplying the original `createdAtMs`) overwrote the row's original insert
timestamp with a fresh `nowMs()`. Since `listMessages` orders by
`createdAtMs`, this silently reordered conversation history and corrupted
downstream oldest-first summarization.

## Fix

Drop `createdAtMs` from the conflict-update `set` clause in
`MemoryStoreLive.js`, mirroring the existing pattern already used by
`setFact`'s `onConflictDoUpdate` (which likewise excludes its own
insert-time-only field from updates). The original insert timestamp is now
preserved across re-saves, while content/threadId/role/runId/etc. still
update as before.

Added a regression test in `packages/memory/tests/store.test.js` that saves
two messages, re-saves the older one with new content and no `createdAtMs`,
and asserts both `listMessages` ordering and the preserved `createdAtMs`
survive.

## Strategy

Built via the **opus-sandwich** strategy.

## Note

This PR will be RESTACKED onto a PR train — its base branch may change after
this is opened.

### DIFF
--- a/packages/memory/src/store/MemoryStoreLive.js
+++ b/packages/memory/src/store/MemoryStoreLive.js
@@ -286,6 +286,11 @@ function makeMemoryStore(db) {
             // crash-resume, deterministic replay, or fork/restore where ids are
             // derived deterministically) must be a safe no-op upsert rather than
             // a UNIQUE-constraint crash. Mirrors setFact's onConflictDoUpdate.
+            // The conflict update deliberately excludes createdAtMs: the ORIGINAL
+            // insert timestamp must survive a re-save so ordering in
+            // listMessages (and downstream oldest-first summarization) stays
+            // stable, even when the re-save omits createdAtMs and would
+            // otherwise fall back to a fresh nowMs().
             yield* writeEffect("memory saveMessage", () => db
                 .insert(smithersMemoryMessages)
                 .values({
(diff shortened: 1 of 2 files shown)


roninjin10:

PR #653, opened 2026-07-08
🐛 fix(vcs): populate WorkspaceInfo.selected in jj workspaceList's -T path

### LINKED ISSUE #534: 🧹 vcs: workspaceList's WorkspaceInfo.selected is always false on modern jj — populate it or drop it
**What happens**
`packages/vcs/src/jj.js:270-300`: when `jj workspace list -T 'name ++ "\n"'` succeeds (every modern jj), each row is returned as `{ name, path: null, selected: false }` — including the currently selected workspace. Only the legacy human-output fallback (lines 282-298) parses the `*` marker and fills `selected` correctly. `tests/jj-workspace.test.js` asserts the always-false behavior, so it's baked in.

**Why it's worth discussing**
`WorkspaceInfo.selected` is exported public API (re-exported from `packages/smithers/src/index.js:191` and documented in the llms bundles). A field that is structurally present but only meaningful on ancient jj versions is a trap for consumers: code that branches on `selected` works in legacy-fallback tests and silently never fires in production.

**Options**
1. Extend the template to capture selection state (e.g. jj's `current_working_copy` template keyword) and populate `selected` in the primary path.
2. Drop `selected` (and `path`, which is also always null in the primary path) from `WorkspaceInfo`, or document them as legacy-only.

roninjin10:
Closes #534

## Root cause

`packages/vcs/src/jj.js`'s `workspaceList()` uses a fast templated `jj workspace list -T ...` path for structured output. The template emitted only the workspace `name`, so the mapped `WorkspaceInfo` rows always had `selected: false` — callers had no way to tell which workspace was actually checked out via this path (the human-output fallback parser handled the `*` marker correctly, but the `-T` fast path did not).

## Fix

- `packages/vcs/src/jj.js`: extend the `-T` template to also emit jj's `current_working_copy()` template keyword (`name ++ "\t" ++ if(target.current_working_copy(), "1", "0")`), then parse that flag into `selected` for each row instead of hardcoding `false`.
- `packages/vcs/src/WorkspaceInfo.ts`: document that `selected` is now populated on both the templated and fallback paths, and that `path` remains unpopulated (no jj template keyword exposes it).
- `packages/vcs/tests/jj-workspace.test.js`: updated assertions to cover the newly-populated `selected` flag on the `-T` path.
- Regenerated the `llms-full.txt` doc bundles (`pnpm docs:llms`) since `docs/reference/vcs-helpers.mdx` changed.

## Strategy

Built via the **opus-sandwich** strategy.

Note: this PR will be **restacked onto a PR train** — its base branch may change as a result.

### DIFF
--- a/apps/cli/docs/llms-full.txt
+++ b/apps/cli/docs/llms-full.txt
@@ -9305,7 +9305,7 @@ type WorkspaceInfo = {
 function workspaceList(cwd?: string): VcsEffect<WorkspaceInfo[]>;
 ```
 
-Prefers template output when supported, falls back to parsing the human-readable `jj workspace list` output.
+Prefers template output when supported, falls back to parsing the human-readable `jj workspace list` output. `selected` reflects the current workspace (populated via jj's `current_working_copy` template keyword on modern jj, or the `*` marker on the legacy fallback). `path` is not currently populated and is always `null`.
 
 ## `workspaceClose(name, opts?)`
 
--- a/docs/llms-core.txt
+++ b/docs/llms-core.txt
@@ -9294,7 +9294,7 @@ type WorkspaceInfo = {
 function workspaceList(cwd?: string): VcsEffect<WorkspaceInfo[]>;
 ```
 
-Prefers template output when supported, falls back to parsing the human-readable `jj workspace list` output.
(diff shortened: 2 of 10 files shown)


roninjin10:

PR #654, opened 2026-07-08
🐛 fix(engine): carry rootDir/maxOutputBytes/taskContext into JSON-repair and schema-retry generate 

### LINKED ISSUE #535: 🐛 engine: JSON-repair follow-up generate() omits rootDir/maxOutputBytes/taskContext
**What happens**
When a task's agent response contains no extractable JSON, the engine sends a follow-up "output ONLY the JSON" repair turn. That call (`packages/engine/src/engine.js:3890-3903`) passes only `options/abortSignal/prompt/timeout/onStdout/onStderr`. The primary generate call (engine.js:3587-3620) passes `rootDir: taskRoot`, `taskContext`, and `maxOutputBytes: toolConfig.maxOutputBytes`; the schema-validation retry call (engine.js:4135-4153) passes `rootDir` and `maxOutputBytes`.

**Why it's wrong**
For CLI agents `rootDir` determines the working directory, so the repair turn runs in the engine process cwd instead of the task worktree (wrong repo state, wrong relative paths, potential writes outside the worktree). Its stdout is also uncapped because `maxOutputBytes` is dropped.

**Expected**
The repair call carries the same execution context as the other two generate calls: `rootDir: taskRoot`, `maxOutputBytes: toolConfig.maxOutputBytes`, and `taskContext` (the schema-retry call is also missing `taskContext` and could pick it up in the same fix).

roninjin10:
Closes #535

## Root cause

`packages/engine/src/engine.js` builds the primary agent `generate()` call
with `rootDir`, `maxOutputBytes`, and a `taskContext` (runId/nodeId/iteration/attempt)
derived from the task descriptor. When the primary response either has no
extractable JSON (JSON-repair follow-up turn) or fails schema validation
(schema-validation retry turn), the engine issues a second `generate()` call
to fix it up — but that follow-up call was built without carrying the same
`rootDir`/`maxOutputBytes`/`taskContext` options through, so repair/retry
turns silently ran with a different execution context (wrong working
directory scoping, no output-size cap, and no task/run identity for tracing)
than the primary turn.

## Approach

Threaded the same `rootDir`, `taskContext`, and `maxOutputBytes` fields from
the primary `generate()` call site into both follow-up call sites in
`engine.js`:
- the JSON-repair follow-up generate call
- the schema-validation retry generate call

Added `packages/engine/tests/json-repair-execution-context.e2e.test.jsx`,
a real-engine e2e (fake in-process agent, no mocks) asserting the repair
and retry calls receive `rootDir`/`maxOutputBytes`/`taskContext` equal to
the primary call's.

Strategy used: opus-sandwich.

Note: this PR will be RESTACKED onto a PR train — its base branch may change
before merge.

### DIFF
--- a/packages/engine/src/engine.js
+++ b/packages/engine/src/engine.js
@@ -4173,6 +4173,14 @@ async function legacyExecuteTask(adapter, db, runId, desc, descriptorMap, inputT
                             options: undefined,
                             abortSignal: taskSignal,
                             prompt: jsonPrompt,
+                            rootDir: taskRoot,
+                            taskContext: {
+                                runId,
+                                nodeId: desc.nodeId,
+                                iteration: desc.iteration,
+                                attempt: attemptNo,
+                            },
+                            maxOutputBytes: toolConfig.maxOutputBytes,
                             timeout: desc.timeoutMs ? { totalMs: desc.timeoutMs } : undefined,
                             onStdout: (text) => {
                                 recordInternalHeartbeat();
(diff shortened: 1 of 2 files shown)


roninjin10:

PR #655, opened 2026-07-08
🐛 fix(engine): track WatchTree waiters per-resolver, not a single slot

### LINKED ISSUE #540: 🐛 engine: hot WatchTree.waitEffect single-resolver slot — concurrent waiters overwrite each other and interrupt cleanup can clear another waiter's resolver
**What happens**
`WatchTree.waitEffect` (`packages/engine/src/hot/watch.js:126-145`) stores exactly one resolver in `this.waitResolve`.

1. Two concurrent waiters: the second registration overwrites the first resolver, so the first fiber never resumes even when changes flush.
2. The Effect.async cancellation cleanup (watch.js:137-141) nulls `this.waitResolve` unconditionally — it does not check the slot still holds *this* waiter's resolver. Interrupting waiter A after waiter B has registered discards B's resolver, so B hangs even after `flush()` fires (changes are only buffered into `changedFiles`).

Any future second consumer of the watch tree (or a race between an interrupted-and-restarted hot-reload loop iteration and its successor) silently hangs the surviving waiter. Latent today — the engine drives a single waiter — but the API gives no hint of the restriction.

**Expected**
Track resolvers per-waiter (a Set) and have each interrupt cleanup remove only its own entry, or explicitly reject a second concurrent `waitEffect`.

roninjin10:
Closes #540

## Root cause

`WatchTree` (`packages/engine/src/hot/watch.js`) stored pending waiters in a
single `waitResolve` slot. When two callers concurrently awaited
`waitEffect()`/`wait()` (e.g. two overlapping hot-reload cycles), the second
`Effect.async` registration overwrote the first waiter's resolver in that
slot. Each `Effect.async` cleanup only knew to null out "the" slot, so
whichever waiter's fiber was interrupted or resolved first could stomp on the
other's pending resolver — leaving the first waiter's fiber permanently
unresolved once a file-change flush landed.

## Fix

Replaced the single `waitResolve` slot with a `waitResolvers` Set. Each
`waitEffect()` call now registers its own resolver closure and removes only
that closure on interruption/cleanup. `flush()` and `close()` iterate a
snapshot of all pending resolvers and resolve every one, instead of resolving
a single slot. Updated the `WatchTree` type surface in
`packages/engine/src/index.d.ts` accordingly, and extended
`packages/engine/tests/hot-watch.test.js` with coverage for two concurrent
waiters both resuming on flush, and for interrupting one waiter leaving the
other's resolver intact.

## Strategy

opus-sandwich

## Note

This PR will be RESTACKED onto a PR train — its base branch may change.

### DIFF
--- a/packages/engine/src/hot/watch.js
+++ b/packages/engine/src/hot/watch.js
@@ -38,7 +38,7 @@ export class WatchTree {
     polling = false;
     pollingDisabled = false;
     currentPollIntervalMs = MIN_POLL_MS;
-    waitResolve = null;
+    waitResolvers = new Set();
     closed = false;
     /**
    * @param {string} rootDir
@@ -81,10 +81,11 @@ export class WatchTree {
             catch { }
         }
         this.watchers = [];
-        // Resolve any pending wait with empty array
-        if (this.waitResolve) {
-            this.waitResolve([]);
-            this.waitResolve = null;
+        // Resolve any pending waits with empty array
+        const resolvers = [...this.waitResolvers];
+        this.waitResolvers.clear();
+        for (const resolveWait of resolvers) {
+            resolveWait([]);
         }
         logInfo("closed hot watch tree", {
             rootDir: this.rootDir,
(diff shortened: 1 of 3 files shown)


roninjin10:

PR #662, opened 2026-07-08
🐛 fix(electric-proxy): report per-stream forwardedBytes, not process total

### LINKED ISSUE #541: 🐛 electric-proxy: `electric.shape.forwarded` reports the process-wide forwardedBytes counter as the closing stream's byte count
**What happens**
When a shape stream releases, the proxy emits `electric.shape.forwarded` with `forwardedBytes: metrics.snapshot().forwardedBytes` (`packages/electric-proxy/src/createSmithersElectricProxy.ts:714`). That counter is the global cumulative total — every stream's `wrapBody` pull increments it (createSmithersElectricProxy.ts:525).

With two concurrent shape streams, each stream's close event reports the sum of both streams' traffic; over a long-lived proxy the field is effectively "bytes since process start", which makes per-shape telemetry (and anything aggregating it) meaningless or double-counted.

**Expected**
Track a per-stream byte count inside `wrapBody` and report that in the event — or rename the field to make clear it is a process total.

roninjin10:
Closes #541

## Root cause

The Electric proxy's `electric.shape.forwarded` event reported `forwardedBytes`
from `metrics.snapshot().forwardedBytes` in
`packages/electric-proxy/src/createSmithersElectricProxy.js` — a single
process-wide counter shared by every concurrent shape stream. When multiple
shapes were streaming concurrently, each stream's "forwarded" event reported
the *summed total across every stream*, not the bytes that particular stream
actually forwarded.

## Fix

`wrapBody()` in `packages/electric-proxy/src/createSmithersElectricProxy.js`
now tracks a local `streamBytes` counter scoped to each individual stream and
passes it into the `release(streamBytes)` hook. The `release` callback that
calls `emitSmithersElectricEvent` (wired up where the proxy handles each
shape request) now sets `forwardedBytes: streamBytes` instead of reading the
shared `metrics.snapshot()`. The JSDoc type for the `release` hook in
`createSmithersElectricProxy.js` and the doc comment on `forwardedBytes` in
`packages/electric-proxy/src/SmithersElectricProxyObserver.ts` were updated
to make clear this is per-stream, not process-wide.

Added a regression test in
`packages/electric-proxy/tests/electric-proxy-hardening.test.ts` that opens
two concurrent shape streams with different byte counts (40 and 80 bytes) and
asserts each stream's `forwarded` event reports its own byte count rather
than the shared/summed total (120).

## Strategy

Built with the **opus-sandwich** strategy.

## Notes

This PR will be **restacked onto a PR train** — its base branch may change
from `main` to an earlier PR in the train.


roninjin10:

PR #756, opened 2026-07-08
🧹 refactor(scheduler): remove unreachable findDescriptor linear-scan fallback

### LINKED ISSUE #546: 🧹 scheduler: findDescriptor's linear-scan fallback is unreachable — decide whether the descriptor map should represent per-iteration descriptors or the fallback should go
**What happens**
`state.descriptors` is only written by `markGraph()` via `descriptorMap()` (`packages/scheduler/src/makeWorkflowSession.js:38-44`), which keys entries by `task.nodeId` — one entry per nodeId, each value's nodeId equal to its key. The fallback in `findDescriptor` (makeWorkflowSession.js:51-58) — `[...state.descriptors.values()].find(c => c.nodeId === nodeId && (iteration == null || c.iteration === iteration))` — can therefore only re-examine the very entry `get(nodeId)` already returned; it can never succeed where the first check failed.

**Why it matters**
- The scan is dead defensive code that reads as if the map could hold multiple iterations of a node; it can't.
- Iteration-qualified lookups for a *previous* iteration of a looped node always return undefined; callers like `unhandledFailureDecision` silently depend on the separate `failureDescriptors` map instead (makeWorkflowSession.js:625-626), while others (`heartbeatTimedOut`, `cacheResolved`) would hit `NODE_NOT_FOUND`/no-op for stale iterations.

**Expected**
A decision: either (a) delete the unreachable fallback and document that `descriptors` holds only the current iteration per node, or (b) key the map by nodeId+iteration if per-iteration lookups are genuinely needed.

roninjin10:
Closes #546

## Root cause

`findDescriptor` in `packages/scheduler/src/makeWorkflowSession.js` did a keyed
`Map` lookup by `nodeId` first, then fell back to a full linear scan
(`[...state.descriptors.values()].find(...)`) over `state.descriptors` if the
keyed lookup missed. But `state.descriptors` is keyed by `nodeId` itself and
holds exactly one entry per node (the current iteration only — no
per-iteration history). That invariant makes the fallback branch dead code:
any entry the linear scan could find would already have been found by the
keyed lookup on the same `nodeId` key. A stale-iteration lookup is intended to
return `undefined` (callers needing a prior failed iteration's descriptor use
the separate `state.failureDescriptors` map instead), so the fallback wasn't
masking a real behavior difference — it was just unreachable, O(n)-scanning
overhead retained on every call site (`unhandledFailureDecision`,
`taskFailed`, `approvalResolved`, `approvalTimedOut`, `timerFired`,
`heartbeatTimedOut`, `cacheResolved`, `cacheMissed`, etc.).

## Approach

Extracted the keyed-lookup-only logic into a standalone,
independently-testable `packages/scheduler/src/findDescriptor.js` module and
dropped the unreachable linear-scan fallback, documenting the
one-entry-per-nodeId invariant in a doc comment. `makeWorkflowSession.js` now
imports `findDescriptor` instead of defining it inline; all call sites are
unchanged. Added `packages/scheduler/tests/findDescriptor.test.js` covering
the keyed-lookup contract (no linear scan, correct entry returned, stale
iteration returns `undefined`).

Strategy used: **fable-sandwich**.

## Note

This PR will be **restacked** onto a PR train — its base branch may change
after this is opened.

### DIFF
--- a/packages/scheduler/src/findDescriptor.js
+++ b/packages/scheduler/src/findDescriptor.js
@@ -1,0 +1,20 @@
+/** @typedef {import("@smithers-orchestrator/graph").TaskDescriptor} TaskDescriptor */
+
+/**
+ * `state.descriptors` holds exactly one entry per nodeId — the CURRENT
+ * iteration — keyed by nodeId. There is no per-iteration history here, so
+ * this is a keyed lookup only: a stale-iteration lookup intentionally
+ * returns undefined. Callers that need a prior failed iteration's
+ * descriptor use the separate `failureDescriptors` map instead.
+ *
+ * @param {Map<string, TaskDescriptor>} descriptors
+ * @param {string} nodeId
+ * @param {number} [iteration]
+ * @returns {TaskDescriptor | undefined}
+ */
+export function findDescriptor(descriptors, nodeId, iteration) {
+    const descriptor = descriptors.get(nodeId);
+    return descriptor && (iteration == null || descriptor.iteration === iteration)
+        ? descriptor
+        : undefined;
+}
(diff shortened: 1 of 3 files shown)


roninjin10:
Superseded by #772 (merged). This PR was part of a stacked chain; its intermediate base branch was deleted during a batch merge, so it auto-closed. The surviving branch carried the cascaded work and was re-targeted to main as #772, which landed on main. No work lost.

PR #769, opened 2026-07-09
Add CursorAgent CLI adapter

ralfboltshauser:
## Summary

- Add a native `CursorAgent` CLI adapter for `cursor-agent` stream JSON output.
- Register Cursor across the public agents and smithers facades, capability registry, CLI surface manifest, docs, and generated docs bundles.
- Add focused Cursor command-building/output-parsing tests and extend CLI capability coverage.
- Update Codex/CodexAgent-adjacent docs and eval fixtures to mention `CursorAgent` where Cursor is now an equivalent supported CLI worker.
- Add Cursor to `smithers init` agent scaffolding and local agent detection (`cursor-agent status --format json` or `CURSOR_API_KEY`).

## Validation

- `bun test --timeout=60000 --max-concurrency=1 packages/agents/tests/cursor-support.test.js packages/agents/tests/cli-capabilities.test.js`
- Direct `generateAgentsTs` check for authenticated `cursor-agent` detection and `CursorAgent` scaffold output
- `bun scripts/generate-llms.ts`
- `git diff --check`

## Known baseline/local-environment limitations

- `./node_modules/.bin/tsc -p packages/agents/tsconfig.json --noEmit` fails on current `main` as well: `packages/agents/src/SdkAgentOptions.ts(10,10): error TS2707: Generic type 'ToolLoopAgentSettings' requires between 0 and 3 type arguments.`
- `node scripts/check-docs.mjs` still has unrelated baseline/environment failures: stale composite docs, optional package import resolution, missing VCS/time-travel deps, and CLI help checks blocked in this partial linked-dependency worktree; Cursor-related CLI agent docs checks pass.
- The full CLI init e2e suite was not rerun end-to-end because this isolated worktree does not have a complete workspace dependency install. `incur` is declared in the repo package manifests; Cursor detection/scaffold generation was validated directly instead.


roninjin10:
Thanks @ralfboltshauser — this is landing. Cursor is now a first-class supported CLI worker (the 15th adapter in the table), including the `smithers init` scaffolding and local detection you wired up.

I rebased your branch onto current `main` (it had moved a fair bit in 16 days) and pushed one polish commit on top of yours. Your two commits keep their authorship and I'm merging with `--rebase`, not squash.

**What I changed and why**

The substantive one: I verified the adapter against the real `cursor-agent` (2026.07.23) rather than the assumed stream shape, and the `tool_call` handling did not match it. The CLI emits

```json
{"type":"tool_call","subtype":"started","call_id":"...","tool_call":{"tool":{"case":"shellToolCall","value":{"args":{...}}}},...}
```

so reading `payload.tool_call_id` / `payload.name` / `payload.args` meant three things went wrong at once: every tool rendered as the literal string `tool`, arguments were always dropped, and `started` and `completed` fell back to two *different* synthetic ids — so the run UI would append an orphan row instead of updating one. That is exactly the bug #1370 had just fixed for `OmpAgent`, so I did not want Cursor shipping with it. Name, args, and failure state now come out of `tool_call.tool`, and failure reads the per-tool result oneof (`success` vs `failure`/`error`/`timeout`/…) instead of a `status` field the CLI never emits. New tests pin the shared-`call_id` parity and the failure path.

Everything else was small:

- **Detector**: `status --format json` only exists on newer `cursor-agent` builds (the one on my box predates it), so I added the credential-file fallback the `claude`/`codex` probes use, plus the real `authSignals` path. I dug the path out of the CLI itself rather than guessing: `~/.cursor/auth.json` on macOS, `$XDG_CONFIG_HOME/cursor/auth.json` on Linux, `%APPDATA%/Cursor/auth.json` on Windows.
- **Routing**: I moved `cursor` to the *end* of each `ROLE_PREFERENCES` list and added it as the last CLI entry in each `TIER_PREFERENCES` order, so it is a live fallback everywhere but never displaces a current leader.
- **Test fixture**: the fake `cursor-agent` binary emitted `message` as a string; the real CLI sends `{role, content:[{type:"text",text}]}`. Fixed so the fixture exercises the real parse path.
- **Pinned expectations you could not have known about**: `capability-registry` and `cli-agent-detection` assert exact adapter lists/counts, and the `init` smoke typecheck has a hand-written `smithers-orchestrator` stub that now needs `CursorAgent`. All three updated.
- **Generated files**: I dropped your versions of `llms-*.txt`, the seeded pack, and the `.d.ts` bundles and regenerated them from source on the rebased tree, so there is no stale-artifact drift. I also dropped `docs/llms-full-v0.28.0.txt` — those versioned bundles are frozen per release and we are past 0.30.0.

**Docs reconciliation.** `main` already had a `docs/agents/cursor.mdx`, but about the opposite direction: Cursor as a *host* driving Smithers over MCP. Your page documented Smithers *spawning* `cursor-agent`. Rather than pick one, the page now opens by naming both directions explicitly and splits into "Drive Smithers from Cursor" and "Run cursor-agent as a Smithers worker", so a reader is never guessing which way round it is. `docs/agents/overview.mdx` and `docs/integrations/cli-agents.mdx` match.

**On your validation notes** — you flagged three things you could not confirm locally. For the record, none reproduce on current `main`: `tsc -p packages/agents/tsconfig.json --noEmit` exits 0, and `check-docs` passes. Those were almost certainly artifacts of the partial install in your worktree, not real baseline breakage. The CLI init e2e suites you could not run do now pass (they needed the three fixture updates above).

**Gates, all on the rebased branch:** `pnpm install --frozen-lockfile`, `typecheck`, `lint`, `check:docs`, `check:llms`, `check:dts`, `check:deps`, `check:sota`, `pnpm -C packages/agents test` (806 pass), `pnpm -C packages/smithers test` (314 pass), and the full `apps/cli` suite (1903 pass, 0 fail).

One thing I deliberately left out of scope: the `ROLE_PREFERENCES` lists are stale against our current routing policy independent of Cursor, so I filed #1384 rather than widen this PR.

roninjin10:
Merged as `4858965a9e` + `d005756d60` (yours) + `3c6eaf70d6` (the polish commit), rebased so your authorship is in main's history.

Post-merge CI note, for the record. Three jobs are red at the merged tip, and none of them are this PR:

- `test (ubuntu-latest, 1, 3)` and `coverage` — `TypeError: Requested module is already fetched` from a react-dom double-fetch. The identical failure is on `7134941c6f`, the commit immediately *before* this one, so it is pre-existing main breakage that someone is already chasing.
- `test-postgres` — `Get "https://registry-1.docker.io/v2/": context deadline exceeded` pulling the postgres image. Docker Hub weather, not code.

The PR's own run was 19/19 green immediately before merge, including those same three jobs, and I re-verified the merged main tip in a clean checkout: `typecheck`, `lint`, `format:check`, `check:docs`, `check:llms`, `check:dts`, `check:deps`, `check:sota`, `packages/agents` (806 pass), and the CLI init/detection suites (57 pass) all green.

One extra rebase happened after my earlier comment: the repo-wide `oxfmt` format pass landed mid-review, so the branch was rebased across it and your files reformatted to the new house style. Nothing semantic changed in that pass.

Thanks again @ralfboltshauser.

PR #1369, opened 2026-07-22
fix: share bounded postgres pools

### LINKED ISSUE #1368: Share bounded PostgreSQL pools across Gateway workflow registrations
Multi-workflow Gateway processes can retain one PostgreSQL client per workflow registration. Three local gateways held 29 idle sessions each and exhausted PostgreSQL connection capacity, causing Smithers inspection failures.

**Proposed change**:
 - Share one bounded PostgreSQL pool per process and normalized connection identity.
 - Default to ten connections with an explicit documented configuration path.
 - Reference-count ownership; close a pool only after its final owner releases it.
 - Preserve transaction safety and failure cleanup.
 - Preserve SQLite and PGlite behavior.
 - Avoid process-global PostgreSQL type-parser mutation.

**Acceptance**:
 - Multiple workflow registrations sharing one PostgreSQL URL share bounded connections.
 - Distinct normalized URLs remain isolated.
 - Final owner cleanup closes the pool.
 - Failure paths release references.
 - Tests cover lifecycle, bounds, transaction handling, and backend compatibility.

**Goal**: upstream a bounded, shared PostgreSQL connection facility for multi-workflow Gateway processes.

mindfather:
Fixes #1368


roninjin10:
Reviewed and polished. Thanks @mindfather - the design here is right: one bounded pool per process per normalized URL, refcounted ownership, transaction affinity preserved by pinning a client from `BEGIN` through `COMMIT`/`ROLLBACK`, and no `pg.types.setTypeParser` process-global mutation. I adopted it as specified in #1368 and pushed one polish commit onto your branch (rebased onto current `main`, your commit first).

**Maintainer changes**

1. **Default pool size 10 -> 16.** Source, option JSDoc, and every docs surface.
2. **Saturation must be loud.** Exceeding the cap previously queued forever behind a leaked client and, once a bound existed, would have surfaced as node-postgres' bare `timeout exceeded when trying to connect`. Acquire now has a bounded wait (`SMITHERS_POSTGRES_ACQUIRE_TIMEOUT_MS`, default 10s) and the timeout is translated into a registered `PG_POOL_SATURATED` `SmithersError` stating the pool identity, the cap and whether it is the default 16, live open/idle/waiting counts, the likely causes, and the exact knob and value to raise it. node-postgres uses a different message for a slow *new* socket, so a network stall is not mislabeled as saturation.

**Review findings fixed**

- **Password leak.** The pool identity is the full URL, so the conflicting-bound `RangeError` printed `postgres://user:secret@host/db` verbatim. Every message now goes through `redactPostgresIdentity()`.
- **Unhandled idle-client error.** Moving from one `pg.Client` to a `pg.Pool` makes an idle-client `error` event (server restart, network drop) likely, and an unhandled one takes down the whole Gateway process. The pool now has a handler; owners still see the failure on their next query.
- **`docs/llms-full-v0.30.0.txt` reverted.** That is a published, frozen snapshot; `pnpm docs:llms` explicitly refuses to rewrite it.

**Verification** - `pnpm typecheck`, `pnpm lint`, `check-docs`, `check-llms` all clean; `packages/smithers` 315/315 and `packages/db` 721/721. Against a live Postgres 16 (`SMITHERS_TEST_PG_URL`) the new `sharedPostgresPool.pg.test.js` covers six registrations collapsing onto one bounded pool with server-side `pg_stat_activity` session counts asserted, real `BEGIN`/`COMMIT` pinning with a concurrent pooled read that must not see the uncommitted row, rollback on mid-transaction close, final-owner close dropping every session, distinct-database isolation, and real saturation; `apps/cli/tests/postgres-roundtrip.e2e.test.js` passes end to end, which is the real proof the pooled connection is a drop-in for `pg.Client` across schema provisioning, migrations, and transactions.

Merging. Closes #1368.

PR #1370, opened 2026-07-24
fix(agents): preserve OMP tool-call args on start/update (parity with PiAgent)

### LINKED ISSUE #1371: Agents: OmpAgent output interpreter drops tool-call args on tool_execution_start/update (parity gap vs PiAgent)
## Bug

The native `OmpAgent` output interpreter (`packages/agents/src/OmpAgent.js`, `createOutputInterpreter`) records each tool call as `{ id, kind, title }` and **drops `payload.args`**. `PiAgent` preserves them as `action.detail.args` on its `tool_execution_start`/`tool_execution_update` handlers (`packages/agents/src/PiAgent.js:323-367`).

Any consumer that reconstructs tool state from interpreter action events therefore loses the arguments when the engine is `omp`, but not when it is `pi` — the two engines emit divergent `action.detail` shapes for the same lifecycle events.

## Symptom

The `todo` tool emits its op payload (`{ op: "init", list: [...] }`, `{ op: "start", task }`, …) as `args` on `tool_execution_start`/`update`. Under `pi` those survive on `action.detail.args`; under `omp` they vanish, so a downstream renderer can only observe *that* a todo call happened, not *what* it did — it cannot rebuild the todo list.

```js
const interpreter = new OmpAgent({ mode: "json" }).createOutputInterpreter();
const args = { op: "init", list: [{ phase: "Investigate", items: ["Map files"] }] };
const action = interpreter
  .onStdoutLine(JSON.stringify({ type: "tool_execution_start", toolName: "todo", toolCallId: "t1", args }))
  .find((e) => e.type === "action");
// action.detail === undefined   (args lost under omp; preserved under pi)
```

## Impact

No current consumer reads `action.detail.args`/`.result` — reconstruction paths (`gateway-ui/src/nodeChat.ts`, `apps/cli/src/chat.js`) key off the PoolAgent-shaped `detail.input`/`detail.output` and fall back to `message`. So the defect is latent today, but it blocks any consumer written against the pi contract from working uniformly across engines. Emitting the same `detail.args` under `omp` closes the parity gap and keeps the engines interchangeable.

## Expected

`OmpAgent`'s interpreter should attach `detail: { args }` on `tool_execution_start`/`update` when `args` is present, mirroring `PiAgent`. The `tool_execution_end` frame carries a `result` rather than `args`; leaving it detail-less is acceptable because no consumer reads `detail.result` from the end/completed frame.

## Verification

Regression test in `packages/agents/tests/omp-support.test.js` asserting `args` survive on start/update and that the end frame stays detail-less; existing argless tool-event interpreter test must remain green.

## Fix

Resolved by #1370.

H3xept:
Closes #1371

## Summary

The native `OmpAgent` output interpreter records each tool call as `{ id, kind, title }` and **drops `payload.args`**. `PiAgent` preserves them as `action.detail.args` (see `PiAgent.js`, the `tool_execution_start`/`tool_execution_update` handlers). Any consumer that reconstructs tool state from OMP events therefore loses the arguments when the engine is `omp` but not when it is `pi`.

Concretely: the `todo` tool emits its op payload (`{ op: "init", list: [...] }`, `{ op: "start", task }`, …) as `args` on `tool_execution_start`/`update`. Under `pi` those survive on `detail.args`; under `omp` they vanish, so a downstream renderer can only see that *a* todo call happened, not what it did — it cannot rebuild the todo list.

## Fix

Attach `detail: { args }` on `tool_execution_start`/`update` when `args` is present, mirroring `PiAgent`. The `tool_execution_end` frame carries a `result` rather than `args`, so it stays detail-less (consumers key off the earlier frames).

```js
const action = { id: asString(payload.toolCallId) ?? name, kind: toolKindFromName(name), title: name };
if (payload.args !== undefined) action.detail = { args: payload.args };
events.push({ type: "action", engine: "omp", phase, entryType: "thought", action, message: name, ok: payload.isError !== true });
```

## Reproduction

Before this change, interpreting an OMP `tool_execution_start` with args yields an action whose `action.detail` is `undefined`:

```js
const interpreter = new OmpAgent({ mode: "json" }).createOutputInterpreter();
const args = { op: "init", list: [{ phase: "Investigate", items: ["Map files"] }] };
const action = interpreter
  .onStdoutLine(JSON.stringify({ type: "tool_execution_start", toolName: "todo", toolCallId: "t1", args }))
  .find((e) => e.type === "action");
// before: action.detail === undefined   (args lost)
// after:  action.detail.args === args    (matches PiAgent)
```

## Tests

Adds a regression test in `packages/agents/tests/omp-support.test.js` asserting args survive on start/update and that the end frame stays detail-less.

```
$ bun test tests/omp-support.test.js
 16 pass  0 fail  42 expect() calls
```

Scope is minimal — one interpreter branch in `OmpAgent.js` plus the test. Behavior is unchanged for tool events without `args` (the existing multi-event interpreter test still passes untouched).


roninjin10:
Reviewed and approved. Thanks @H3xept — the diagnosis and the parity claim both check out against `PiAgent.js:323-367`.

**Apologies are owed here.** While this PR was open, an automated `opus5-bug-sweep` run of ours independently reimplemented the same fix and landed it on main as 162f8d143b, and closed #1371 against that commit. That should not have happened: your PR came first and carries the authorship. Our maintainer policy is that a contributor's PR is reviewed and merged, never reimplemented, and the sweep bypassed it. We're tightening the sweep's pre-flight to check for an open PR referencing the issue before it picks up work.

To make it right, this PR is merged rather than closed, so your commit lands in history as the fix.

**What I changed on your branch**

- Merged main in and resolved the collision. I kept main's form of the same fix: it is equivalent to yours plus an explicit `phase !== "completed"` guard, so a hypothetical end frame that echoed `args` still stays detail-less rather than relying on omp never sending them.
- Dropped the near-duplicate of the test main already had, and added a cross-engine test that drives the *same* payloads through both the `OmpAgent` and `PiAgent` interpreters and asserts `action.detail.args` matches. That verifies the parity claim against PiAgent's real behavior instead of a hand-copied shape, which is what a "parity with PiAgent" fix should be pinned by. Confirmed red without the fix.
- Regenerated `packages/agents/src/index.d.ts`. Unrelated to your change: 162f8d143b altered the interpreter's action shape without refreshing the checked-in declarations, so the `check-dts` gate (`ci.yml:184`) has been failing on main since it landed. That is what the first red run on this PR was; it is fixed here.

`packages/agents`: 797 pass, 0 fail. typecheck, lint, check-dts all clean.

Issue #1372, opened 2026-07-24
OmpAgent should default streaming runs to a persistent RPC session (not one-shot --print)


H3xept:
## Problem

The native `OmpAgent` defaults to a one-shot `--print --mode json` execution
whenever Smithers streams events (`resolveMode` returns `"json"` for
`options.onEvent`). RPC — the persistent session protocol the interactive omp TUI
uses — is opt-in only via `mode: "rpc"`.

For a coding/tool-using agent this is the wrong default: the one-shot structurally
prevents the multi-turn tool loop the agent exists for, and it does not match the
omp TUI mental model. Observed on a real multi-step task:

| | one-shot (`--print --mode json`) | RPC session |
|---|---|---|
| turns | single-shot | 62 |
| tool calls | ~0 | 8 |
| wall time | ~9s | ~50s |
| behavior | single Q&A then exit | persistent session, like the TUI |

## Proposal

Default streaming runs to `--mode rpc`. Keep `mode: "json"` as an explicit
one-shot opt-out for genuinely single-turn work (classification/extraction), and
fall back to one-shot json automatically when the caller passes file arguments
(RPC rejects them today).

## Tradeoffs

- (+) Spawned omp agents behave like the TUI; multi-step tool work actually loops.
- (−) A persistent RPC session costs more tokens/latency than a one-shot; cheap
  single-turn Task nodes should opt out via `mode: "json"`.

Draft PR implementing this follows.

roninjin10:
Closed by #1373. Thanks @H3xept for proposing, implementing, and validating the persistent RPC default; the merged PR preserves the explicit `mode: "json"` opt-out and file-argument fallback, with maintainer polish for usage, structured output, environment parity, and process teardown.

PR #1373, opened 2026-07-24
feat(agents): default omp streaming runs to a persistent RPC session

H3xept:
Refs #1372

## Summary

Flip the native `OmpAgent` streaming default from a one-shot `--print --mode json`
to a persistent `--mode rpc` session (the protocol the omp TUI uses), so spawned
omp coding agents actually run their multi-turn tool loop instead of answering
once and exiting.

- `resolveMode`: streaming (`options.onEvent`) now defaults to `"rpc"`.
- Falls back to `"json"` when file arguments are present (`buildArgs` rejects
  `rpc` + files), so that path is unchanged.
- Explicit `mode: "json"` remains the one-shot opt-out for cheap single-turn work.
- `generate()` already routes to the RPC path for `mode === "rpc"`, so no other
  wiring changes.

## Evidence

Real multi-step task: one-shot = single-shot / ~0 tool calls / ~9s; RPC = 62
turns / 8 tool calls / ~50s (persistent session, like the TUI).

## Tests

Adds cases in `packages/agents/tests/omp-support.test.js`: streaming defaults to
RPC; streaming + file args falls back to json; explicit `mode:"json"` opt-out
stays one-shot. Full `packages/agents` suite passes.

## Note

Opened as **draft** — this changes a default and its blast radius (cost/latency
for single-shot Task nodes) is a maintainer call. Marking ready once reviewed.

### DIFF
--- a/apps/cli/docs/llms-full.txt
+++ b/apps/cli/docs/llms-full.txt
@@ -18125,11 +18125,30 @@ import { OmpAgent } from "smithers-orchestrator";
 const omp = new OmpAgent({
   provider: "openai",
   model: "gpt-5.6-terra",
-  mode: "json",
   cwd: process.cwd(),
   autoApprove: true,
 });
 ```
+
+### Mode defaults
+
+When `mode` is omitted, Smithers picks it from the call. A streaming run (any
+`<Task>`, or a `generate()` that passes `onEvent`) defaults to
+`rpc`: a persistent session, the same protocol the interactive OMP TUI speaks, so
+the agent runs its full multi-turn tool loop instead of answering once and exiting.
+Everything else defaults to `text`.
+
+Set `mode: "json"` to opt back into a one-shot `--print --mode json` execution. That
+is the cheaper choice for genuinely single-turn work such as classification or
+extraction, where a persistent session only costs extra tokens and latency. A
+streaming run that passes file arguments falls back to one-shot `json`
+automatically, because RPC mode rejects files.
+
+| `mode` | Streaming run | Non-streaming run |
+| --- | --- | --- |
+| omitted (default) | `rpc`, or `json` when file arguments are passed | `text` |
+| `"json"` | one-shot `--print --mode json` | one-shot `--print --mode json` |
+| `"rpc"` | persistent session; file arguments are rejected | persistent session |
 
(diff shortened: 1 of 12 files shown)


roninjin10:
Maintainer decision: **adopting the proposal as written.** Streaming runs default to `--mode rpc`, `mode: "json"` stays the explicit one-shot opt-out for single-turn work, and file arguments fall back to one-shot json automatically. The reasoning in #1372 is right — a one-shot structurally prevents the tool loop a coding agent exists for, and the 62-turn/8-tool-call vs single-shot evidence settles it. Thanks @H3xept.

`resolveMode` is correct as written: explicit `mode` wins, non-streaming is untouched, and the `files` fallback lands before RPC's rejection in `buildArgs`. I reviewed the areas the default flip makes load-bearing:

- **Event-stream parity** — fine. The RPC path feeds `runRpcCommandEffect`'s `onJsonEvent` through the *same* `createOutputInterpreter()`, so tool-call, text-delta, started and completed events are identical between modes. Usage flows through `result.usage` (extracted from `event.usage`/`message.usage`) into the shared `buildGenerateResult`, and structured output rides `.text` the same way, so `outputSchema` parsing is unaffected.
- **Session teardown** — already solid, no leaked omp processes. `runRpcCommandEffect` spawns `detached: true` and kills the whole process *group*: `sendAbort()` writes an omp `abort` frame, `terminateChild()` does SIGTERM then SIGKILL after 250ms, the abort signal is wired to `kill()`, and the Effect finalizer SIGKILLs the group on interruption. That covers Task cancellation and run end (cf. #1378).

**What I pushed on top**

One real regression the flip would otherwise have introduced. `generate()`'s RPC branch built its env as `{ ...process.env, ...this.env, ...credentials }`, which means it ignored `inheritEnv: false` and dropped the `SMITHERS_*` identifiers that BaseCliAgent's one-shot path forwards via `taskContextEnv`. Latent while RPC was opt-in; with RPC as the streaming default, every streaming omp task would silently lose its run/node context and break `smithers ask-human` from inside the agent. Extracted the env build into `resolveRpcEnv()` mirroring the one-shot path, with a test pinning both behaviors (verified red without the fix).

Also regenerated `packages/agents/src/index.d.ts` for the `resolveMode` signature change (the `check-dts` gate would have caught it), and documented the mode defaults — neither docs page actually stated them before, so there is now a "Mode defaults" section with a mode/streaming matrix in the CLI-agents page plus a pointer from the OmpAgent reference entry, and `mode: "json"` is dropped from the constructor example since it is no longer the useful default to show. llms bundles regenerated; check-docs and check-llms pass.

Your three tests cover the spec exactly (streaming⇒rpc, explicit json opt-out, file-args fallback). `packages/agents`: 797 pass, 0 fail; typecheck and lint clean.

Merging once #1370 lands and this rebases onto it.

roninjin10:
Additional RPC-default polish is now pushed as 13524ce418:

- normalize RPC usage into the AI SDK shape so Task token telemetry does not read snake_case usage as zero;
- parse RPC answer JSON into `result.output`, matching one-shot json and preserving output-schema behavior;
- terminate the live RPC child on every error-settlement path, with POSIX process-group kill and a Windows direct-child fallback;
- pin tool-event parity, usage, structured output, SIGTERM/SIGKILL teardown, plus the required default/opt-out/file-fallback cases.

Verified after rebasing on #1370: packages/agents 799 pass / 0 fail; typecheck; lint; check-dts; check:docs; check:llms.

PR #1375, opened 2026-07-24
🐛 fix(vibe-audit): green up main; .smithers typecheck + UI-architecture gate

### LINKED ISSUE #1374: 🐛 CI: main is red — .smithers typecheck + UI-architecture gate fail on the landed vibe-audit demo
## Bug

`main` is red. The vibe-audit demo commit (`02b0b6575`, ✨ feat(demoday-site): vibe-audit demo arc) landed three defects that fail two independent CI jobs — the `typecheck` matrix (`pnpm --filter ./.smithers typecheck`) and the `test` shard-0 gate suite (`node scripts/check-ui-architecture.mjs`). None are covered by the demoday-site build, so they only surface in the repo-wide gates.

### 1. Workflow input schema passed to the wrong `createSmithers` argument
`createSmithers(schemas, opts)` takes the input schema as the `input` KEY of the first (schemas) argument; the second argument is `CreateSmithersOptions` (`{ readableName?, description?, alertPolicy? }`) and has no `input` field (`packages/smithers/src/CreateSmithersOptions.ts`). `.smithers/workflows/vibe-audit.tsx:39-50` instead passes `{ input: inputSchema }` as the second argument, so the schema is dropped and `ctx.input` stays `unknown`:

```
workflows/vibe-audit.tsx:49:5 - error TS2353: Object literal may only specify known properties, and 'input' does not exist in type 'CreateSmithersOptions'.
workflows/vibe-audit.tsx:90:25 - error TS18046: 'ctx.input' is of type 'unknown'.
```

### 2. UI compares run-tree status to a non-existent `"finished"` state
`useGatewayRunTree` returns `status: NodeStatus`, and `NodeStatus = "ok" | "running" | "queued" | "failed" | "waiting" | "cancelled"` (`packages/gateway-react/src/sync/useGatewayRunTree.ts:8`) — the completed tone is `"ok"`, there is no `"finished"`. `.smithers/ui/vibe-audit.tsx:193` compares against `"finished"`:

```
ui/vibe-audit.tsx:193:16 - error TS2367: This comparison appears to be unintentional because the types 'NodeStatus' and '"finished"' have no overlap.
```

The same stale `"finished"` literal drives the per-node parked/recovered logic in `StrategyCard` (`:67-68`, `:95`); those do not error (node status is typed `string`) but never match the real `"ok"`, so the parked-on-quota → recovered beat silently never renders.

### 3. Pack-UI architecture violation: inline `style` props
`.smithers/ui/vibe-audit.tsx` renders inline `style={...}` props throughout, tripping the pack-UI guard:

```
UI architecture guard failed:
- New architecture violation: pack-ui-style-prop :: .smithers/ui/vibe-audit.tsx renders a style prop
```

The baseline is an exact-ratchet and its policy note explicitly forbids allowlisting new violations ("do not add entries for new violations"), so the file must be brought into conformance rather than baselined.

## Symptom

- `typecheck` job (ubuntu + windows): `pnpm --filter ./.smithers typecheck` exits 2 — `Found 3 errors in 2 files`.
- `test` job, Linux shard 0 / Windows shard 1: the `check-ui-architecture` gate fails, failing the whole gate step.

## Impact

Every push/PR to `main` fails CI, blocking merges. The vibe-audit demo workflow also cannot read its own input (`ctx.input.repo`) and its UI cannot detect run completion, so the demo is functionally broken beyond the red build.

## Expected

- `.smithers` typecheck is clean.
- The UI-architecture guard passes with no new allowlist entries.
- vibe-audit's workflow types `ctx.input` from `inputSchema`, and its UI reflects real run/node completion (`"ok"`).

## Verification

- `pnpm --filter ./.smithers typecheck` → 0 errors.
- `node scripts/check-ui-architecture.mjs` → "guard passed".
- `pnpm lint` and `pnpm -r typecheck` remain green.

## Fix

1. `.smithers/workflows/vibe-audit.tsx`: move `input: inputSchema` into the first `createSmithers` schemas object and drop the second argument.
2. `.smithers/ui/vibe-audit.tsx`: replace `"finished"` comparisons with `"ok"`; convert all inline `style` props to `className`s backed by CSS injected through `<WorkflowUiStyles extra={...} />` (no raw `<style>` tag, no `style` prop), matching the conformant pack-UI pattern.

H3xept:
## What

`main` is red. The vibe-audit demo commit (`02b0b6575`) landed three defects that fail two independent CI jobs:

| Job | Gate | Cause |
|---|---|---|
| `typecheck` | `pnpm --filter ./.smithers typecheck` | 3 TS errors in the vibe-audit files |
| `test` (Linux shard 0 / Windows shard 1) | `node scripts/check-ui-architecture.mjs` | new un-allowlisted `pack-ui-style-prop` violation |

Fixes #1374.

## Changes

**`.smithers/workflows/vibe-audit.tsx`**
- Move `input: inputSchema` into the first `createSmithers` schemas argument and drop the invalid second argument. `createSmithers(schemas, opts)`'s second arg is `CreateSmithersOptions` (`{ readableName?, description?, alertPolicy? }`) and has no `input` field, so the schema was dropped and `ctx.input` stayed `unknown`. Fixes `TS2353` (`:49`) and `TS18046` (`:90`).

**`.smithers/ui/vibe-audit.tsx`**
- Replace `status === "finished"` with `status === "ok"`. `useGatewayRunTree` returns `NodeStatus = "ok" | "running" | "queued" | "failed" | "waiting" | "cancelled"` — the completed tone is `"ok"`, so the comparison had no overlap (`TS2367` at `:193`). The same stale literal drove `StrategyCard`'s parked/recovered logic and never matched real state; corrected to `"ok"`.
- Convert every inline `style` prop to a `className` backed by a CSS block injected through `<WorkflowUiStyles extra={…} />` (which emits the `<style>` from inside gateway-ui), routed through the styleguide theme tokens so the UI follows the active dark/light theme. No raw `<style>` tag and no `style` prop remain, so the file conforms to the pack-UI guard without a new baseline entry (the exact-ratchet baseline forbids allowlisting new violations).
- Drop the now-dead `theme` / `useState` imports; inline a single-use helper per repo TS rules.

## Verification

- `pnpm --filter ./.smithers typecheck` → 0 errors
- `pnpm -r typecheck`, `pnpm typecheck:examples`, `pnpm typecheck:evals` → pass
- `node scripts/check-ui-architecture.mjs` → *guard passed* (no new allowlist entries)
- `node --test scripts/check-ui-architecture.test.mjs` → 69/0
- `pnpm lint` → 0 errors

---

## Update — remaining red jobs from the same landing

Two further jobs were red on `main` (both since the last-green commit `7be658c8d`), traced to the demoday/vibe-audit landing and fixed here:

- **`.smithers` test** — `workflow-component-inventory.test.ts` ("shipped workflow and component test ownership") requires every discoverable workflow to have exactly one focused owner suite wired into the package `test` script. `vibe-audit.tsx` shipped unowned. Added `tests/vibe-audit-workflow.test.tsx` (schema defaults + the parallel-scans → `deps-audit` retry beat → dedupe/triage/report graph), registered it in `workflowOwners`, and added it to the `.smithers` `test` script. Exported `inputSchema` from the workflow so the suite can assert its defaults.
- **`apps/demoday-site` test** (+ `coverage`) — `worker.test.ts` reads the built deck from `dist/`, but CI runs `bun test` without a build, so the suite errored at load on both OSes and in the coverage job. Prefixed the test script with `vite build` (offline: the narration mp3s + `manifest.json` are committed under `public/` and copied into `dist/`). The compound script also makes the coverage runner skip the package cleanly (it only instruments direct `bun test` scripts) — appropriate for a static pitch deck.

Note: the `windows (…, 4, 4)` shard flakes independently (it passed on `main` and `.smithers`/demoday changes here can't affect that Windows shard); a re-run should clear it.

### DIFF
--- a/.smithers/tests/vibe-audit.test.tsx
+++ b/.smithers/tests/vibe-audit.test.tsx
@@ -6,6 +6,11 @@ import workflow from "../workflows/vibe-audit.tsx";
 
 const workflowPath = `${import.meta.dir}/../workflows/vibe-audit.tsx`;
 
+type XmlNode = { tag?: string; props?: Record<string, unknown>; children?: XmlNode[] };
+
+const childIds = (node: XmlNode | undefined): string[] =>
+  (node?.children ?? []).map((child) => String(child.props?.id ?? ""));
+
 describe("vibe-audit workflow", () => {
   test("registers its input schema so ctx.input carries the repo", () => {
     // createSmithers takes the input schema as the `input` KEY of its schemas
@@ -15,6 +20,8 @@ describe("vibe-audit workflow", () => {
 
     expect(inputSchema).toBeDefined();
     expect(inputSchema?.parse({})).toEqual({ repo: "acme/payments-api" });
+    expect(inputSchema?.parse({ repo: "smithersai/payments-api" })).toEqual({ repo: "smithersai/payments-api" });
+    expect((inputSchema as z.ZodType | undefined)?.safeParse({ repo: 42 }).success).toBe(false);
   });
 
   test("runs four strategies in parallel ahead of the dedupe/triage/report chain", async () => {
@@ -31,6 +38,31 @@ describe("vibe-audit workflow", () => {
     ]);
     // The deck's parked-on-quota beat needs a retry budget on deps-audit.
     expect(frame.tasks.find((task) => task.nodeId === "deps-audit")?.retries).toBe(2);
+  });
+
+  test("fans the four scans out under <Parallel>, then funnels dedupe -> triage -> report", async () => {
+    // `frame.tasks` is a flat list, so it reads the same whether the scans are
+    // concurrent or sequential. Assert the rendered graph itself: only the tree
+    // shape pins the demo's fan-out/funnel beat.
+    const frame = await renderWorkflow(workflow, { workflowPath, input: { repo: "acme/payments-api" } });
+    const root = JSON.parse(frame.toXml()) as XmlNode;
+
+    expect(root.tag).toBe("smithers:workflow");
+    expect(root.props?.name).toBe("vibe-audit");
+
+    const sequence = root.children?.[0];
+    expect(sequence?.tag).toBe("smithers:sequence");
+
+    const parallel = sequence?.children?.[0];
+    expect(parallel?.tag).toBe("smithers:parallel");
+    expect(childIds(parallel)).toEqual(["injection-scan", "auth-review", "secrets-scan", "deps-audit"]);
+
+    // The consolidation pipeline runs strictly after the parallel scans, in order.
+    expect((sequence?.children ?? []).slice(1).map((child) => String(child.props?.id ?? ""))).toEqual([
+      "dedupe",
+      "triage",
+      "report",
+    ]);
   });
 });
 
--- a/apps/demoday-site/package.json
+++ b/apps/demoday-site/package.json
@@ -7,7 +7,7 @@
   "scripts": {
     "dev": "vite",
     "build": "vite build",
-    "test": "bun test tests",
+    "test": "vite build && bun test tests",
     "typecheck": "tsc -p tsconfig.json --noEmit",
     "deploy": "wrangler deploy",
     "deploy:dry-run": "wrangler deploy --dry-run",


roninjin10:
Thank you for this, @H3xept — and I'm sorry about the timing. Full transparency on what happened:

Our own automated bug sweep raced you and landed a fix for the same red main this afternoon (#1374, c5ba780ae8) while your PR was open. That's entirely on our side, not yours — you diagnosed it correctly and independently. I did **not** want to close this unmerged, so I've rebased your branch onto current `main` and kept the parts of your work that main still needs. Both commits remain authored by you.

**What landed from your work**

1. **`apps/demoday-site` — build the deck before its worker test.** This is a real defect our sweep missed entirely. `tests/worker.test.ts` reads `dist/index.html` at module load, `dist/` isn't committed, and CI runs `bun test` without a build — so the suite errored at load on both OSes and in the coverage job. I reproduced it on a clean checkout of current `main` before applying your fix:
   ```
   ENOENT: no such file or directory, open '.../apps/demoday-site/dist/index.html'
   0 pass, 1 fail, 1 error
   ```
   With your `vite build && bun test tests`: **7 pass, 0 fail**, and it works offline exactly as you noted (the mp3s + `manifest.json` are committed under `public/`). Good catch.

2. **The stronger vibe-audit graph assertion.** Your test suite asserted the rendered XML tree — that the four scans are children of `smithers:parallel` and that dedupe → triage → report follow strictly after. Ours only checked the flat `frame.tasks` list, which reads *identically* whether the scans are concurrent or sequential. Your approach is strictly better, so I folded those assertions into the existing owner suite (`.smithers/tests/vibe-audit.test.tsx`) rather than adding a second owner file, which `workflow-component-inventory` forbids. I proved it earns its place by flipping `<Parallel>` to `<Sequence>` in the workflow: your new assertion goes red while all three pre-existing tests stay green.
   ```
   Expected: "smithers:parallel"
   Received: "smithers:sequence"
   ```
   I also kept your tighter input-schema checks (explicit override + rejecting a wrong-typed `repo`).

**What I dropped, and why** — only because `main` already had an equivalent or slightly stronger version, not because anything was wrong with yours:

- `.smithers/workflows/vibe-audit.tsx` — your `createSmithers({ input, ... })` single-argument fix is *character-for-character* what our sweep landed. Same diagnosis, same fix. (`export`ing `inputSchema` is no longer needed since the folded test reads it off the default export.)
- `.smithers/ui/vibe-audit.tsx` — your `"finished"` → `"ok"` diagnosis was right. Main's version reaches the same place and additionally pins it with a typed `const COMPLETED: NodeStatus = "ok"`, so a future stale literal is a compile error rather than a silently-never-matching comparison, plus it extracts `quotaBeat` for unit testing and routes the CSS through styleguide tokens without hex fallbacks. Your style-prop → `WorkflowUiStyles` conversion was the correct call and main does the same thing.

Your write-up of the trap (`createSmithersOptions` has no `input` field, so the schema is silently dropped and `ctx.input` stays `unknown`; `NodeStatus` has no `"finished"` member) is the clearest statement of it anywhere — it's preserved verbatim as the comment above the input-schema test.

**Verified on the rebased branch** (current `main`): `node scripts/check-ui-architecture.mjs` passed with no new allowlist entries · `node --test scripts/check-ui-architecture.test.mjs` 69/0 · `.smithers` `tsc --noEmit` 0 errors · `.smithers` vibe-audit + workflow-component-inventory 8 pass/0 fail · `apps/demoday-site` 7 pass/0 fail from a clean `dist/` · `.smithers` lint 0 errors.

Merging as soon as CI is green. Thanks for taking the time to fix someone else's red main — the demoday build fix in particular is something we'd have kept tripping over.

roninjin10:
Merged as 6ccb8321a025399c63f5b41a301d31eb4934354c — thank you again, @H3xept.

One correction to my comment above, because you got raced a *second* time and you deserve to know exactly what happened:

While this PR sat waiting on CI, another of our automated sweeps landed 64c87bb515 ("💚 fix(ci): build demoday dist before its worker tests"), which contains the **identical** one-liner you wrote:

```
-    "test": "bun test tests",
+    "test": "vite build && bun test tests",
```

Same fix, character for character, arrived at independently ~45 minutes before I merged. `main` enforces linear history, so GitHub rebase-merged this PR — and your demoday commit rebased to *empty* (the content was already there) and was dropped. That's why only one commit carries your name on `main`:

- **6ccb8321a0** — `✅ test(vibe-audit): pin the fan-out/funnel graph shape, not just the task list`, authored by you.

To be unambiguous about credit: **you diagnosed and fixed the demoday build defect first and independently.** Our sweep just happened to push the same line while your PR was in review queue. The fix is on `main` and it's yours as much as it is 64c87bb515's — the commit ledger just doesn't show it, and that's an artifact of our tooling racing you twice, not of your work being redundant.

What's on `main` from you now:
- The `<Parallel>` graph-shape assertion, which closed a real hole — the previous test only checked the flat `frame.tasks` list and could not distinguish concurrent from sequential scans.
- The tighter input-schema checks (explicit override + wrong-typed `repo` rejection).
- The demoday `vite build` fix, in content.

Verified green at the merged tip (6ccb8321a0): `.smithers` vibe-audit + workflow-component-inventory 8 pass/0 fail · `.smithers` `tsc --noEmit` clean · `check-ui-architecture` passed · `apps/demoday-site` 7 pass/0 fail from a clean `dist/`.

Note for the record: `coverage` and `test (ubuntu-latest, 1, 3)` were red on this PR, but they are red on `main` too (run 30181947593 at 369de285d3, the base) for an unrelated pre-existing `apps/cli` failure — four `TypeError: Requested module is already fetched` module-registry errors. Nothing to do with your changes. Your branch actually *fixed* `test (windows-latest, 2, 4)`, which was red on that same main run and passes here.

Sorry for making you fix our red main and then racing you on it twice. Genuinely appreciated — please do send more.

PR #1376, opened 2026-07-24
feat(tooling): adopt oxfmt as the repo formatter (builds on @paolohajinz75's .editorconfig)

paolohajinz75:
Adopts **oxfmt** (the Oxc formatter) as this repository's formatter. This started as @paolohajinz75's `.editorconfig` — the seed of it, and still the first commit here — and grew into the tooling that actually enforces what that file describes.

An `.editorconfig` is a request to your editor; nothing verifies it. Since the repo already lints with **oxlint**, its sibling formatter is the natural pick: one toolchain, one install, and no lint/format rule overlap to arbitrate.

### Configuration, measured rather than guessed

`.oxfmtrc.json` is `useTabs: false`, `tabWidth: 2`, `printWidth: 120`, `endOfLine: lf`. Those values were chosen by formatting the whole scope with each candidate and comparing the resulting diff:

| setting | files touched |
|---|---|
| **spaces, width 2, printWidth 120** | **3,325** |
| spaces, width 4, printWidth 120 | 3,901 |
| tabs, printWidth 120 | 3,994 |
| spaces, width 2, printWidth 100 | 3,452 |
| spaces, width 2, printWidth 80 | 3,746 |

The repo was already space-dominant (3,355 of 3,622 JS/TS files), so this is the setting that disturbs the least existing code.

### Scope of the formatting pass

Formatted: every first-party `.js/.jsx/.mjs/.cjs/.ts/.tsx/.mts/.cts` file under `packages/*/{src,tests}`, `apps/`, `scripts/` and `.smithers/` — the same code `pnpm lint` already covers.

Deliberately excluded:

- everything in `.gitignore`, which oxfmt honours (`reference/`, `.opentui_temp/`, `dist/`, build output);
- `*.d.ts` and `*.generated.*` — build output whose layout a generator owns and a gate (`check:dts`, `check:sota`) re-runs and diffs;
- the other languages oxfmt can format. Pointing it at a directory would rewrite `docs/**/*.mdx`, the `.smithers` prompts and the `docs/llms-*.txt` bundles, so `scripts/format.mjs` globs on JS/TS extensions instead.

### Commands and CI

- `pnpm format` writes, `pnpm format:check` verifies.
- `pnpm lint` and `pnpm lint:fix` now chain oxlint **and** oxfmt, so the everyday command runs both tools.
- CI gains a `pnpm format:check` step next to `pnpm lint`.

`scripts/format.mjs` mirrors `scripts/lint.mjs`: it expands `packages/*/{src,tests}` in Node and spawns oxfmt directly rather than letting a shell see the globs, because cmd.exe and PowerShell do not expand them — the same trap that once broke oxlint on Windows CI.

### What the pass shook loose

A repo-wide reformat is a good fuzzer for gates that assert on source *text*. Four were pinned to a layout rather than to an API, and are fixed at the assumption (not by exempting files):

- `check-docs` compared exact substrings, and read `smithersErrorDefinitions` and the Gateway RPC registry with a hardcoded four-space indent. It now normalizes whitespace before comparing, so a needle matches whether the source prints on one line or wrapped.
- `check-docs` also had six assertions of the form `[FILE]` with the needle missing. They passed only because `String.includes(undefined)` searches for the literal `"undefined"`, which that guide contains. Removed, and a non-string needle now throws.
- `check-dependency-boundaries` ran its dynamic-import regex over raw file text, so an import quoted *inside a string literal* counted as a real dependency. It now blanks string and template literals first.
- `sota.test.ts` and `packages/db/tests/published-sandbox-types.test.js` pinned JSX props and class methods with `[^\n]` / a literal four-space indent. Both now allow the declaration to wrap.

### Blame hygiene

The mechanical pass is its own commit and is listed in `.git-blame-ignore-revs`. Configure it locally with:

```sh
git config blame.ignoreRevsFile .git-blame-ignore-revs
```

### Verification

`pnpm typecheck`, `pnpm lint`, `pnpm format:check`, `pnpm check:dts`, `pnpm check:docs`, `pnpm check:llms`, `pnpm check:deps`, `pnpm check:sota`, `pnpm check:effect`, `pnpm check:ui-architecture`, `pnpm check:db-access`, and the full `pnpm -r test` workspace suite.

Thanks @paolohajinz75 — your PR is the reason this exists.


.editorconfig
```diff
@@ -0,0 +1,19 @@
+# EditorConfig â€” https://editorconfig.org
```
H3xept: remove non ascii character

H3xept (changes requested):

.editorconfig
roninjin10:
```suggestion
# EditorConfig https://editorconfig.org
```

roninjin10:
We should also add biome prettier or something too since this is mostly a typescript project

roninjin10:
Thanks @paolohajinz75 — merging this, and it grew a fair bit on the way in.

Your `.editorconfig` was right about the problem: contributors were getting different indentation because nothing in the repo said what the indentation should be. The thing is, an `.editorconfig` is a request to your editor — nothing verifies it, so the repo would keep drifting. So I used your PR as the reason to finally pick a formatter and wire it up, and kept your commit as the first one in the branch.

The choice is **oxfmt**, the Oxc formatter, because the repo already lints with its sibling **oxlint**: one toolchain, one install, no lint/format rules to arbitrate between.

Two things I changed in your file, both to match what the repo actually looks like:

- the `[*]` fallback said `indent_size = 4`. The repo is overwhelmingly 2-space (3,355 of 3,622 JS/TS files are space-indented), so 4 was wrong for anything that did not hit the more specific glob;
- the JS/TS glob omitted `.mjs` and `.mdx`, and we have a lot of both (`scripts/*.mjs`, all of `docs/`). Added those plus `.cjs`/`.mts`/`.cts`.

Your `end_of_line = lf` stayed exactly as you wrote it — `.gitattributes` already forces LF because Windows CI and the generated bundle diffs are line-ending sensitive, so you were matching the repo there.

The formatter settings were measured rather than guessed: formatting the whole tree with each candidate and comparing diffs gave 3,325 files touched at (spaces, width 2, printWidth 120) versus 3,901 at width 4 and 3,994 with tabs.

A nice side effect: a repo-wide reformat turns out to be an excellent fuzzer for gates that assert on source *text*. It shook loose six assertions across `check-docs`, `check-dependency-boundaries` and three test files that were pinned to a particular layout rather than to an API — including six `check-docs` assertions that had lost their needle entirely and were passing only because `String.includes(undefined)` searches for the literal `"undefined"`. Those are all fixed at the assumption.

Appreciate you opening this.

roninjin10:
Merged — thanks again @paolohajinz75. Your commit is the base of the branch and stayed authored by you (rebase-merge, not squash), so `git log` and `git blame` both still credit you for the `.editorconfig`.

`pnpm format` and `pnpm format:check` are live now, `pnpm lint` runs oxlint and oxfmt together, and CI gates formatting on both Linux and Windows.

PR #1428, opened 2026-07-27
🐛 fix(server): preserve approval decision payloads across HTTP routes

### LINKED ISSUE #1421: Gateway approve route drops select-mode decision payloads ({selected, notes} lost)
aomi labs (0.26.1) built a side-channel POST /decide loopback endpoint to work around it (their console.ts:76,116). Decision payloads submitted through the gateway approve route lose {selected, notes}. Source: aomi deep-dive 2026-07-27, production user.

roninjin10:
Closes #1421

Threaded approval decisions through legacy routes, added shared validation, regression tests, and API docs.

---
Filed from the aomi labs deep-dive (2026-07-27).
Pipeline: Fable investigated, Codex Luna implemented, Fable + Codex Sol both approved in a review loop, a human reviewed at the landing gate.


roninjin10:
## Automated code review

Reviewed 4 file(s) in the diff against `main`. **3 minor**.

### `packages/server/src/approvalDecision.js`:101 — minor/correctness (confirmed)

Fail closed when a `select` or `rank` request has no valid parsed options. `parseApprovalRequest` filters malformed option entries, so a JavaScript workflow such as `options: [{ key: "safe" }]` produces an empty `request.options`; both membership checks are then skipped and an arbitrary selection or ranking is accepted and persisted as workflow output. Return `INVALID_REQUEST` for option-bearing modes with no valid options, or preserve an option-parse error.

```suggestion-ish
function validateApprovalDecision(request, decision) {
  if ((request.mode === "select" || request.mode === "rank") && request.options.length === 0) {
    return { ok: false, code: "INVALID_REQUEST", message: `${request.mode} approval has no valid options` };
  }
  if (request.mode === "select") {
```

### `packages/server/src/approvalDecision.js`:113 — minor/correctness (confirmed)

Reject mixed-type ranking arrays instead of sanitizing only the validation copy. `parseStringArray` drops invalid members, so `{ ranked: ["canary", 7] }` passes validation when `canary` is allowed, but `approveNode` serializes the original decision unchanged. The stored `decisionJson` then violates the declared `string[]` contract and different consumers can observe different rankings.

```suggestion-ish
const payload = asObject(decision);
    const ranked = payload?.ranked;
    if (
      !Array.isArray(ranked) ||
      ranked.length === 0 ||
      ranked.some((value) => typeof value !== "string")
    ) {
      return { ok: false, code: "INVALID_REQUEST", message: "rank approvals require decision.ranked to be a non-empty string array" };
    }
```

### `packages/server/src/index.js`:27 — minor/correctness (confirmed)

Re-export this binding. The package wildcard exposes `approvalDecision.js` at runtime while mapping its types to `index.d.ts`; because this file only imports the binding, TypeScript consumers of that subpath cannot import `approvalDecision`.

```suggestion-ish
import { approvalDecision } from "./approvalDecision.js";
export { approvalDecision };
```


<sub>Generated by `smithers review` (Codex sol), findings adversarially verified.</sub>

PR #1498, opened 2026-08-06
🐛 fix(tui): restore the parens on the Tree inspector's empty logs-tab label

roninjin10:
## Root cause

The Tree inspector's **logs** tab renders `no log events for this node` — unparenthesized, unlike every other inspector placeholder in the monitor (`(no output)`, `(no output yet)`, `(failed, no output)`, and `RunEventLog`'s own `(no log events)` default).

The scaffold (`262d84af7c`) rendered the string inline:

```tsx
<text fg="#444444">{"  (no log events for this node)"}</text>
```

Extracting the shared `packages/tui-ui` `RunEventLog` leaf moved it into an `emptyLabel` prop and **dropped the parentheses in the move**. Nothing caught it: the only assertion on the rendered text lives in the PTY monitor e2e, and that suite is `describe.skipIf(ZMUXD == null || AGENT_HARNESS)` — CI has no zmux, so it only ever runs on a dev box with the daemon installed.

This is a **product bug**, not a test bug. Users see the inconsistent placeholder in `smithers monitor` / `smithers up --interactive` every time they press `[2]` on a node that has no events yet (a parked root, a pending task).

## Evidence — actual vs expected pane content

Headless render of the real `NodeInspectorView` (`packages/tui/tests/tree-render.test.tsx`), before the fix:

```
Expected to contain: "(no log events for this node)"
Received: " output  logs  diff  props
  no log events for this node
```

Real PTY capture through zmux, **before** the fix — the test's regex `/\(no log events for this node\)|^\s+\[\d+\] /` matches neither alternative, so it polls to its ~33s timeout:

```
● e2e-approval-probe  run-mshua4fr  00:02  [paused]
 ▾ ⏸ e2e-approval-probe              │ output  logs  diff  props
  · ⏸ Approve E2E gated task         │  no log events for this node
```

The same capture **after** the fix:

```
● e2e-approval-probe  run-mshu9qfb  00:02  [paused]
 ▾ ⏸ e2e-approval-probe              │ output  logs  diff  props
  · ⏸ Approve E2E gated task         │  (no log events for this node)
```

Reverting only the one-string fix reproduces the failure and re-applying it fixes it — direct causation, no other variable moved.

## Changes

- `packages/tui/src/modes/TreeMode.tsx` — restore the parens on `emptyLabel`, with a comment recording why the extraction lost them.
- `packages/tui/tests/tree-render.test.tsx` — **new CI coverage.** The suite already asserted the *non-empty* logs tab (`[11]`, `tool.use`) but never the empty one, which is exactly the state that regressed. This is the cheap non-PTY closure of the gap and runs on every box.
- `apps/cli/tests/tui-monitor-zmux.e2e.test.js` — on failure, throw with the captured pane grid instead of a bare `expect(...).toBe(true)`. Same strictness (the test still fails if the logs tab breaks), but the next person gets the actual terminal content in the failure message instead of `received: false`. Matches the existing `startParkedMonitor` precedent in the same file.

## Verification

`packages/tui`: **448 pass / 0 fail** across 30 files. The new test fails without the fix and passes with it.

PTY, run twice back to back against a real zmuxd + real gateway + real parked run:

```
===== RUN 1 =====
  · ⏸ Approve E2E gated task         │  (no log events for this node)
(pass) 2 renders the logs tab placeholder [5941.33ms]
 1 pass  0 fail
===== RUN 2 =====
  · ⏸ Approve E2E gated task         │  (no log events for this node)
(pass) 2 renders the logs tab placeholder [5958.50ms]
 1 pass  0 fail
```

`pnpm typecheck` and `pnpm lint` are clean for the three touched files (the only pre-existing failures on this tree are a duplicated `@tanstack/query-core` in `packages/gateway-client` and Prettier drift in unrelated foreign WIP).

## Note: a separate, unrelated blocker on this dev tree

The three tests in `tui-monitor-zmux.e2e.test.js` currently all fail *earlier* than the logs-tab assertion — at `startParkedMonitor`, with `monitor never rendered the parked gate`. That is **not** this bug and not caused by this change. The shared checkout has accumulated **166** workflows in `.smithers/workflows`, and the gateway fires `Promise.all(discoveredWorkflows.map(loadWorkflowByKey))` right after publishing liveness (`apps/cli/src/index.js`). Those 166 concurrent module evaluations starve the event loop, so the socket accepts but answers nothing: a probe gateway burned **5+ minutes of CPU without serving a single request** at load average 30. Running the identical PTY flow against an isolated single-workflow workspace completes in **~6s**. Worth its own issue — the gateway's "liveness before workflow import" guarantee does not hold under workspace scale — but deliberately out of scope here.


roninjin10:
Superseded: the fix landed on main as aa5324cdd3 ("🐛 fix(tui): restore the empty logs placeholder") together with the headless tree-render regression test (packages/tui/tests/tree-render.test.tsx:95) and a 0.34.0 changelog entry. This branch additionally carries ~21 unrelated commits from a stale base (Trellis submodule work, nanocodex scripts) that conflict with main in 13 files, so closing rather than rebasing. Verified on origin/main: TreeMode.tsx renders emptyLabel="(no log events for this node)".

PR #1499, opened 2026-08-06
🐛 fix(testing): stop tests leaking scratch dirs into $TMPDIR

roninjin10:
## The leak

Test and e2e helpers created per-test scratch dirs with `mkdtemp`/`mkdtempSync` in `$TMPDIR` and never removed them.

- **2026-08-05:** 30,132 `smithers-*` dirs = 35 GB, disk-full emergency (4 GB free of 926 GB).
- **2026-08-06:** filled the volume again; a release dry-run failed with 32 `ENOSPC` errors in the engine suite. After a manual reclaim it regrew by ~3,900 dirs within hours.

Observed families on disk (prefix identifies the responsible helper): generic unprefixed `smithers-*` (1598, includes the ~40 MB `smithers-migrate-*/smithers.db` offenders), `smithers-sandbox-execute-` (108), `smithers-fallback-agents-` (80), `smithers-sandbox-bundle-` (52), `smithers-identity-claude-` (50), `smithers-snap-` (35), `smithers-fake-docker-bin-` (32), `smithers-bubblewrap-` (20), `smithers-identity-codex-` (15), `smithers-tgc-cov-` (11), `smithers-fake-cli-` (10), `smithers-conventions-test-` (10).

## The fix

**`packages/testing/src/cleanup/tempDir.ts`** — one helper, used everywhere:

```ts
makeTempDir(prefix?)   // -> { path, cleanup, [Symbol.dispose] }
makeTempDirPath(prefix?)  // -> string, for call sites that only needed the path
cleanupTempDirs()      // drains the registry
trackedTempDirs()      // what is currently outstanding (used by the guard)
```

Every directory it creates goes into a module-level registry, drained by:

1. `using dir = makeTempDir(...)` / `dir.cleanup()` — immediate, per-use.
2. an `afterAll` armed on the file that created the directory — runs even when the test throws.
3. `process.on("exit")` — for plain `node`/`bun run` consumers.

Only paths the helper created are ever removed. Live state such as the running gateway's `$TMPDIR/smithers-gateway` is never in the registry and is never touched.

### Two bun behaviours shaped the design (both documented in the helper)

- **`bun test <dir>` runs every file in one process with one module registry.** A hook registered when the helper module first loads only covers the file that imported it first — the rest of the suite keeps leaking. The drain is therefore armed from `makeTempDir` (whenever the registry goes empty → non-empty), which self-scopes to the calling file.
- **Registering a bun hook from inside an `AsyncLocalStorage` context hangs the enclosing test for 5 s** (`a beforeEach/afterEach hook timed out…`, bun 1.4 canary; reproducible with a 10-line `als.run(...)` repro). `execute.test.js` now allocates its sandbox root *before* entering `withTaskRuntime` rather than inside the callback.

## Regression guard

`packages/testing/tests/tempDir.test.ts` runs a leak fixture and a real suite (`packages/sandbox/tests/sandbox-bundle.test.js`) in child `bun test` processes with `$TMPDIR` pointed at a directory the guard itself created, then asserts no `smithers-*` entries survive. Because the child's `$TMPDIR` is private, concurrent test runners cannot perturb it — no global counting. It also asserts the child actually ran tests, so a filter mistake cannot make the guard pass vacuously.

Verified it fails when a raw `mkdtemp` is reintroduced into the fixture.

## Measured before / after

"Before" is the accumulated count in `$TMPDIR` (2,464 `smithers-*` dirs at the start of this work). "After" is a full run of each package with an isolated, empty `$TMPDIR`, counting what remains:

| Package / suite | before (on disk) | after (leaked by a full run) |
| --- | --- | --- |
| sandbox (execute, bundle, transport-runners, utils, isolation, heartbeat) | 108 + 52 + 32 + 20 + … | **0** |
| agents (fallback-agents) | 80 + 8 | **0** |
| apps/cli (identity-claude / identity-codex / fake-cli) | 50 + 15 + 10 | **0** |
| db (snap, provenance) | 35 | **0** |
| integrations (tgc-cov, gh, gh-cov, telegram, tg-approval, lin-cov) | 11 + 22 | **0** |
| agent-eliza (conventions-test, loader-cov, fm-test) | 10 + 11 | **0** |
| smithers migrate suites (`smithers-migrate-*`, ~40 MB each) | 120 | **0** |
| generic unprefixed `smithers-*` (helpers.js) | 1598 | **0** |
| engine (cache-root, logdir, vcs-git, name/nameless) — the suite that hit ENOSPC | 16 per run | **0** |
| vcs (find-vcs-root) | — | **0** |
| components / time-travel / xstate / react-reconciler | — | **0** |

The only entry the engine run still leaves is `$TMPDIR/smithers-durability`, the product's fixed-name durability gap spool (`packages/engine/src/durabilityGapSpool.js`) — live run state, deliberately untouched.

## Product (non-test) call sites — audited, no change needed

`apps/cli/src/share.js`, `apps/cli/src/ask.js`, `packages/agents/src/KimiAgent.js`, `packages/server/src/gatewayRoutes/getRunDiff.js`, and `packages/testing/src/coverWorkflow.ts` all remove their directories in a `finally` or expose a `cleanup()`. `packages/smithers/src/external/create-external-smithers.js` and `packages/vcs/src/jj.js` create long-lived state that must survive; they are intentionally left alone.

## Not converted

The remaining `mkdtemp` call sites in `packages/agents/tests` (and similar) already clean up after themselves — they left nothing behind in the isolated-`$TMPDIR` runs, so they were not churned.

## Gates

- `pnpm lint` — clean for every file touched here.
- `pnpm typecheck` — no errors in any file touched here (4 pre-existing errors remain in `packages/gateway-client`, caused by duplicate `@tanstack/query-core` versions from unrelated in-flight dependency work).
- Package tests all pass: sandbox (200), agents (908), db (736), integrations (264), agent-eliza (94), testing (171), components (739), engine (1149), time-travel (299), xstate (65), react-reconciler (88), vcs (95), the smithers migrate/backend suites (77), and the touched apps/cli suites (18).

---

## Review follow-up (17fc34d)

Two gaps found reviewing the above, fixed in a follow-up commit on this branch:

1. **`makeTempDir` now namespaces every directory `smithers-*`.** `makeWorkspace(name)` in the smithers migrate/backend suites passes a bare suite name, so the helper was producing `$TMPDIR/migrate-reverse-leftover-sqlite-…` instead of `smithers-migrate-reverse-leftover-sqlite-…`. Those are the ~40 MB PGlite workspaces — the heaviest family in the outage — and under a bare name they drop out of the `smithers-*` census this whole investigation was measured by, so a future regression there would be invisible to the very measurement that found the bug. The prefix is now prepended when absent, which cannot be forgotten at a call site. Covered by a new unit test.
2. **The regression guard now asserts on every entry in the isolated `$TMPDIR`, not just `smithers-*` ones.** A clean suite leaves that directory completely empty (verified empirically), so the prefix filter only blinded the guard to the regression it most needs to catch: a new raw `mkdtemp` under some other prefix.

Also corrected a comment in `migrateStoreKit.js` that claimed the registry's process-exit sweep reclaims workspaces when a suite dies mid-run. `bun test` never runs `process.on("exit")` handlers (verified), so the per-test drain is the only backstop there; the sweep helps plain `node`/`bun run` consumers only.

### Re-verified after the follow-up

`$TMPDIR` held 2,613 `smithers-*` dirs before the run and **2,613 after** — a net delta of **0 new directories** across full package suites for `sandbox`, `agents` (908 tests), `db` (736), `integrations` (264), `agent-eliza` (94), `testing` (172), `vcs` (95), and `smithers`, plus the migrated `engine` e2e and `apps/cli` suites. Previously-unnamespaced `migrate-*` dirs: 2 pre-existing on disk, **0 new**.

`pnpm lint` and `node scripts/format.mjs --check` are clean for every file touched. The one remaining `packages/smithers` test failure (`published type stubs and JSX runtime typecheck`) and the `pnpm typecheck` errors are the pre-existing `packages/gateway-client` duplicate `@tanstack/query-core` (5.101.2 vs 5.101.4) problem from unrelated in-flight dependency work — unchanged by this branch.


roninjin10:
Superseded: the entire fix is on main — packages/testing/src/cleanup/tempDir.ts is byte-identical to this branch, and the call-site migration landed too (main has fewer raw-mkdtemp sites than this branch covers). The only raw mkdtemp files on main not covered here (e2e/faults/case32-generic-agent-checkpoint-kill-resume.test.ts, e2e/runtime/run-node-cli.mjs) were created after this branch diverged, so they need a fresh follow-up, not this PR. The branch also carries ~23 unrelated stale-base commits that conflict with main. Closing.

PR #1526, opened 2026-08-14
Guard oneshot preflight commits while runs are active

roninjin10:
References bug 01kzzaqfx1g9qaefqxrderdz4m.

## Summary
- query non-terminal workspace runs for the same canonical cwd and classify them through PID-verified run state
- prevent auto preflight from committing or taking ownership of dirty paths while live or waiting runs are active, with an ID-bearing warning and fail-closed store errors
- add --preflight-force-commit as an explicit override, retain the safer wait/worktree recommendation, and propagate it through detached and resume launches
- document the contract and regenerate llms bundles

## Race review
The check runs after working-copy assessment and before dispatch; detached launches repeat it in the admitted child, which closes the parent-to-child spawn window. Store inspection is read-only, and inspection failures fail closed. Finished, stale-heartbeat, dead-PID, orphaned, and cancel-pending rows do not block. A run created after the final read is the remaining store-level TOCTOU boundary, so the warning recommends an isolated worktree when launches may overlap.

## Verification
- bun test --timeout=120000 apps/cli/tests/oneshot-preflight.test.js apps/cli/tests/oneshot.test.js
- pnpm typecheck
- pnpm lint
- pnpm docs:llms


roninjin10:
Closing as a duplicate of #1525. Both PRs implement defect A (bug 01kzzaqfx1g9qaefqxrderdz4m) and they conflict in four files (`apps/cli/src/index.js`, `apps/cli/src/oneshot-preflight.js`, `apps/cli/src/oneshot/buildOneshotChildArgs.js`, `apps/cli/tests/oneshot-preflight.test.js`), so only one can land.

#1525 is the one to keep:

- Its holding-state set includes `stale`. `deriveRunState` reports `stale` when the heartbeat lapsed but the owner PID is alive or unproven, which is a saturated live run. This PR's `ACTIVE_RUN_STATES` omits `stale`, so a busy run in the same cwd would not block the commit. That is the exact scenario the bug describes.
- It keeps one named export per file (`findActiveRunsInCwd`, `runConfigRootDir`, `isWorkingCopyHoldingRunState`) instead of adding a third export to the shared preflight module.
- It reuses the existing `--preflight` enum for the `force-commit` escape hatch instead of adding a second flag.
- It degrades softly per status query, so one failed `listRuns` call does not hide the runs the other queries found.

The branch stays pushed, so this can be reopened if #1525 turns out to be the wrong call.

PR #1588, opened 2026-08-16
fix(db): one refcounted sqlite connection per file per process (does NOT fix #1577)

### LINKED ISSUE #1577: CI ubuntu shard 3: recurring SQLite wedge fails jumpToFrame/devtools/webhook server tests
## Symptom
Since the first red main run (feat(nanocodex) #1515, 2026-08-14T05:09Z), `test (ubuntu-latest, 3, 3)` fails the same four packages/server tests on nearly every run:

- `jumpToFrame auth > pause path fails with RewindFailed when task does not terminate within 10s` (times out; response code is `Busy` instead of `RewindFailed`)
- `gateway devtools serializer warnings and audit recovery` (both tests; 20-30s timeouts, previously instant SQLITE_BUSY at `createSmithers` open)
- `Gateway webhook ingestion > rejects invalid webhook signatures and records rejection metrics` (webhook counters never appear in /metrics within 10s)

## Evidence
- The cluster appears in every ubuntu shard-3 job since run 31772092690 and does not reproduce locally.
- In the jumpToFrame test the rewind's audit writes and lease release fail sequentially, each after a full 30s `busy_timeout`, with `SQLITE_PROTOCOL` ("locking protocol"). Timestamps show exact 30s gaps (see run 31867035170, shard 3 log).
- The lease heartbeat dies during those stalls, so the rollback path reports the fence-loss code `Busy` and masks the original `RewindFailed`. The test then fails on the code assertion.
- PR #1570 made `SQLITE_PROTOCOL` retryable and set `busy_timeout` before the `journal_mode` switch. That converted the instant failures into 20-30s stalls, confirming the lock is held for tens of seconds rather than transient.

## Hypothesis
Two SQLite connections to the same database file inside one bun test process: the `createSmithers` bun:sqlite `Database` and the effect-sql client used by gateway register-time recovery, event-bridge polling, and devtools pollers (the detached jobs #1558 drains on close). bun:sqlite calls are synchronous, so a writer waiting on `busy_timeout` blocks the event loop and prevents the same-process holder from ever releasing; the wedge only resolves when the timeout expires. Timing shifted when #1515 changed the shard weight distribution and co-scheduled these suites.

## Suggested directions
- Ensure one process uses one connection per database file, or route the secondary writers through the primary connection's write queue.
- Alternatively, drop `busy_timeout` for same-process secondary connections and rely on the async retry schedule so the event loop keeps running.

## Not the cause
- Runner infra flakes (apt-get hang, chocolatey) also appear on other shards but are unrelated.
- The webhook metrics assertion races were made poll-based in #1570; the test still fails on shard 3 because the process is wedged, not because of the assertion.

roninjin10:
> **Read this first: this PR does _not_ fix #1577.** It is left open for a decision — see "Negative result" below. The `Closes #1577` claim in the original description was wrong and has been removed.

## Negative result

CI shard 3 on this branch still fails the exact same four tests, with the same 30-second-stall signature, with the fix in place:

- `gateway devtools serializer warnings and audit recovery` (both tests)
- `Gateway webhook ingestion > rejects invalid webhook signatures and records rejection metrics`
- `jumpToFrame auth > pause path fails with RewindFailed when task does not terminate within 10s`

So the "two in-process connections to one file" hypothesis in #1577 is **not** the cause, or not the whole cause.

The local Linux-container evidence that pointed the other way does not hold up: running the full `packages/server` suite in an ubuntu container, the unfixed baseline failed the webhook test and the fixed build passed it — but both runs also carried environmental failures (no Chromium in the image) and ran under heavy host load, so that single flip was almost certainly noise rather than signal. I am recording it as a false positive.

## What this branch still contains, on its own merits

A process-wide, refcounted `bun:sqlite` connection registry keyed by resolved absolute path (`@smthrs/db/sqliteConnectionRegistry`), anchored on a `Symbol.for` global so a duplicated module instance cannot hand out a second connection. Every in-process open routes through it: `createSmithers`, `openDurableSqliteDatabase`, `createBuilderDb`, `openSmithersStore`, `create-external-smithers`, `openInlineWorkflowStore`.

`close()` releases one reference and the real `sqlite3_close` runs at zero, so existing close semantics — including `Gateway.close()` as an I/O barrier — keep working; repeated closes from one holder are no-ops. `:memory:` and `file:` URIs are never shared. An entry whose file was deleted, or whose handle was closed behind the registry's back, is evicted rather than reused. Pragmas apply only on the open that actually creates the connection. `scripts/check-no-direct-db-access.mjs` and the CLI boundary test both ratchet down accordingly.

Whether that invariant is worth having independently of #1577 is the maintainer's call. It is a real tightening (one connection per file per process is a defensible rule on its own), but it touches every SQLite open path, so it should not land just because it is written.

## Gates

`packages/db` 817 pass · `packages/smithers` 333 pass · `packages/engine` pass · `packages/server` `server.test.js` + `gateway-pty-hijack.test.jsx` 58 pass · new `packages/db/tests/sqlite-connection-registry.test.js` 6 pass · `check-no-direct-db-access` · `pnpm lint` · `pnpm format:check` · per-package typechecks.

### DIFF
--- a/apps/cli/src/openInlineWorkflowStore.js
+++ b/apps/cli/src/openInlineWorkflowStore.js
@@ -3,15 +3,21 @@ import { SmithersError } from "@smthrs/errors";
 
 /** @param {string} cwd @param {Record<string, import("zod").ZodTypeAny>} schemas */
 export async function openInlineWorkflowStore(cwd, schemas) {
-  const [{ Database }, { drizzle }, { sqliteTable, text }, { zodToTable }, { syncZodTableSchema }, { camelToSnake }] =
-    await Promise.all([
-      import("bun:sqlite"),
-      import("drizzle-orm/bun-sqlite"),
-      import("drizzle-orm/sqlite-core"),
-      import("@smthrs/db/zodToTable"),
-      import("@smthrs/db/zodToCreateTableSQL"),
-      import("@smthrs/db/utils/camelToSnake"),
-    ]);
+  const [
+    { acquireSqliteConnection },
+    { drizzle },
+    { sqliteTable, text },
+    { zodToTable },
+    { syncZodTableSchema },
+    { camelToSnake },
+  ] = await Promise.all([
+    import("@smthrs/db/sqliteConnectionRegistry"),
+    import("drizzle-orm/bun-sqlite"),
+    import("drizzle-orm/sqlite-core"),
+    import("@smthrs/db/zodToTable"),
+    import("@smthrs/db/zodToCreateTableSQL"),
+    import("@smthrs/db/utils/camelToSnake"),
+  ]);
   const { findSmithersAnchorDir } = await import("smthrs/findSmithersAnchorDir");
   const { resolveSmithersBackendPreference } = await import("smthrs/resolveSmithersBackendChoice");
   const anchorDir = findSmithersAnchorDir(cwd);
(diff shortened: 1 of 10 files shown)


roninjin10:
Closing this. Its hypothesis is eliminated and the real fix has landed.

This PR routed all six in-process `bun:sqlite` opens through a refcounted registry keyed by resolved
absolute path, so one process held exactly one connection per file. Shard 3 failed identically, which
is why it was parked rather than merged.

#1577 is now understood: the wedge was **resource exhaustion from the number of simultaneously open
handles**, not duplicate connections. The fix (PR #1617) is a tracked-fixture teardown sweep that
takes peak live handles from 189 to 55, and it is confirmed across two independent shard-3 runs —
failing tests went 5 to 1, and the `SQLITE_PROTOCOL` / `locking protocol` signature is gone from both.

The prerequisite that made it possible was also real: `createSmithers` had no reachable `close()`,
because `create.js` wired its `closeDb` closure only to `process.once("exit")`. Callers could not
release a handle even if they wanted to. That is fixed in #1617 as well.

Nothing here is worth salvaging on top of that, so this closes rather than rebases.

roninjin10:
Closing this. Its hypothesis is eliminated and the real fix has landed.

This PR routed all six in-process `bun:sqlite` opens through a refcounted registry keyed by resolved
absolute path, so one process held exactly one connection per file. Shard 3 failed identically, which
is why it was parked rather than merged.

#1577 is now understood: the wedge was **resource exhaustion from the number of simultaneously open
handles**, not duplicate connections. The fix (PR #1617) is a tracked-fixture teardown sweep that
takes peak live handles from 189 to 55, and it is confirmed across two independent shard-3 runs —
failing tests went 5 to 1, and the `SQLITE_PROTOCOL` / `locking protocol` signature is gone from both.

The prerequisite that made it possible was also real: `createSmithers` had no reachable `close()`,
because `create.js` wired its `closeDb` closure only to `process.once("exit")`. Callers could not
release a handle even if they wanted to. That is fixed in #1617 as well.

Nothing here is worth salvaging on top of that, so this closes rather than rebases.

PR #1592, opened 2026-08-16
🐛 fix(engine): stop a dead resume session from consuming retry budget

roninjin10:
## The defect

A resumed agent CLI session whose persisted id is dead raises `AGENT_SESSION_LOST` with `discardResumeSession: true`. No work is dispatched — the agent throws the id away and the next attempt opens a fresh session, which the error message itself promises. The engine nonetheless counted it as an ordinary failure, so it consumed the task's retry budget.

Quota-limited attempts were already exempt for exactly this reason ("the run pauses until the quota resets and then retries as if the attempt never occurred"). Session losses are the same class of bookkeeping failure and were not exempt.

Observed on a real 7-lane run: one task's attempts were

1. heartbeat timeout (a real failure)
2. quota (correctly preserved)
3. **dead resume id (consumed a retry)**
4. quota (correctly preserved)
5. **dead resume id (consumed a retry)**

and the task died as "Attempt 5 of 3" having suffered exactly one real failure. Stale resume ids are routine for anyone running long tasks: the CLI prunes old conversations, machines restart, ids age out.

## The fix

A new `isDiscardedSessionAttempt` predicate, applied at the three sites that count retry-consuming attempts:

- `effect/retry-state.js` — `failureCount` for `stampDurableRetryState`
- `effect/workflow-bridge.js` — `retryConsumingFailures` in `classifyTaskAttempt`
- `engine.js` — the retry-rung counter and both `retryConsumingFailedAttempts` filters

Quota-specific park/resume logic is deliberately untouched: a session loss must never park a run waiting on a provider reset.

A session that broke even though the attempt started **fresh** still consumes the budget, so the task fails over to the next agent in the chain rather than retrying something that cannot succeed. The kimi classifier already reported that distinction via `freshSessionFailure`; the claude classifier now does too, with a message that says which case it is.

## Verification

- New `packages/engine/tests/discarded-session-retry-budget.test.js` (9 tests) pins the bug: it **fails on unmodified main** (`failureCount` 3 instead of 1) and passes with the fix. Covers meta-vs-error flag sources, malformed payloads, the fresh-session exclusion, and that real failures still exhaust the budget.
- `packages/engine`: 54 pass / 0 fail across the new file plus `engine-internals` and `failure-streak`.
- `packages/agents`: 42 pass / 0 fail across `classifyQuotaError` and `claude-support`.
- `oxlint` clean on every changed file (one pre-existing unrelated warning at `BaseCliAgent.js:622`).

### DIFF
--- a/packages/agents/tests/classifySessionLoss.test.js
+++ b/packages/agents/tests/classifySessionLoss.test.js
@@ -83,6 +83,33 @@ describe("classifySessionLoss", () => {
     expect(err.message).not.toContain("Retry will start a fresh session");
   });
 
+  test("claude crash on a FRESH session is flagged so it counts as an agent failure, not a fixable resume", () => {
+    const err = classifySessionLoss(
+      "claude",
+      "No conversation found with session ID: 68b187d0-a325-4384-a248-b2a0e6edbd90",
+      "",
+      false,
+    );
+    expect(err).not.toBeNull();
+    expect(err.code).toBe("AGENT_SESSION_LOST");
+    expect(err.details.freshSessionFailure).toBe(true);
+    // Still discard the broken id: the heartbeat may have captured it.
+    expect(err.details.discardResumeSession).toBe(true);
+    expect(err.message).toContain("FRESH session");
+    expect(err.message).not.toContain("Retry will start a fresh session");
+  });
+
+  test("a resumed claude session is flagged as recoverable so it does not consume retry budget", () => {
+    const err = classifySessionLoss(
+      "claude",
+      "No conversation found with session ID: 68b187d0-a325-4384-a248-b2a0e6edbd90",
+      "",
+      true,
+    );
+    expect(err.details.freshSessionFailure).toBe(false);
+    expect(err.message).toContain("Retry will start a fresh session");
+  });
+
   test("the claude pattern on a different CLI is NOT session loss", () => {
     expect(classifySessionLoss("codex", "No conversation found with session ID: abc12345", "")).toBeNull();
   });
--- a/packages/engine/src/effect/isDiscardedSessionAttempt.js
+++ b/packages/engine/src/effect/isDiscardedSessionAttempt.js
@@ -1,0 +1,41 @@
+/**
+ * A resumed agent session whose persisted id is dead fails the attempt before
+ * any work is dispatched: the agent discards the id and the next attempt opens
+ * a fresh session. Like a quota-limited attempt, such a failure must never
+ * consume the task's retry budget — otherwise a long task whose session ids go
+ * stale (the CLI pruned the conversation, the machine restarted, the id aged
+ * out) exhausts its budget on session bookkeeping instead of real failures.
+ *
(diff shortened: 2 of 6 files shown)


roninjin10:
CI triage: the `test (ubuntu-latest, 3, 3)` failure is pre-existing on `main`, not caused by this branch.

`main` at `c5ed0ac3` ([run 31977492801](https://github.com/smithersai/smithers/actions/runs/31977492801)) fails the same shard with the same tests:

- `jumpToFrame auth > pause path fails with RewindFailed when task does not terminate within 10s` (30s timeout)
- `gateway devtools serializer warnings and audit recovery > a too-deep snapshot tree drives the serializer onWarning on both routes` (30s timeout)

All three failures on this branch are timeout-bound gateway/devtools tests that touch neither retry-budget accounting nor session classification. The packages this PR changes are green: `packages/engine` 54 pass / 0 fail (new suite plus `engine-internals` and `failure-streak`), `packages/agents` 42 pass / 0 fail.

The earlier `typecheck` red on the first push was `format:check` flagging one unformatted file; that is fixed and typecheck now passes.

roninjin10:
Two notes from hitting this exact failure from the other direction today.

**1. The `BaseCliAgent.js` hunk here looks stale-based.** `classifySessionLoss` on current `main` already carries `hadResumeSession` / `freshSessionFailure` on the claude branch, with the same message split this diff adds. Worth rebasing so the review is just the engine change.

**2. The engine change and #1597 are complementary, and neither is sufficient alone.**

This PR makes a discarded-session attempt stop consuming retry budget. #1597 adds the missing **codex** branch to `classifySessionLoss` — codex had none, so a codex dead resume never got classified as session loss at all and therefore never reached the exemption this PR adds.

The failure I hit five times today: a codex stream drops before the rollout is recorded, the thread id is captured but never persisted, and every subsequent attempt resumes it and gets

```
thread/resume failed: no rollout found for thread id <id> (code -32600)
```

With only this PR, codex still isn't classified, so the budget exemption never fires. With only #1597, codex is classified but the attempts still count. Together they do what the error message has always promised.

They do touch the same two files, so whichever lands second will need a trivial rebase — happy to take that on #1597.

roninjin10:
Rebased onto current `main` (was 19 commits behind).

One conflict, resolved in favor of upstream: `main` has independently landed the same claude fresh-session split in `classifySessionLoss`, so that half of this branch is now redundant and I took main's wording verbatim.

**The core of the fix is still needed.** `isDiscardedSessionAttempt` appears zero times in main's `packages/engine/src/effect/retry-state.js` — upstream now *classifies* a fresh-session failure correctly but still lets a dead resume session consume the task's retry budget. That is the defect this PR closes, at the three sites that count retry-consuming attempts.

Re-verified on the rebased branch: engine suite 7/7, agents `classifySessionLoss` 16/16, `format:check` clean.

roninjin10:
Closing: this targets the 0.x engine, which no longer exists on main. The 1.0 engine (packages/engine-store) has its own resume and retry-budget model; if the same failure reproduces there, open a fresh issue against it.

PR #1593, opened 2026-08-16
fix(agents): boot the parent-death watchdog outside the agent's cwd

### LINKED ISSUE #1546: review action: codex file reviews intermittently fail with 'Cannot find module unist-util-visit-parents/do-not-use-color' under Bun
The `review` check (smithers review action) fails intermittently across PRs. When it fails, every codex file-review node dies instantly with:

```
error: Cannot find module 'unist-util-visit-parents/do-not-use-color' from '/home/runner/.bun/install/cache/unist-util-visit-parents@6.0.2@@@1/lib/index.js'
Bun v1.3.13 (Linux x64)
```

Example failing run: https://github.com/smithersai/smithers/actions/runs/31817917663 (PR #1543, 2026-08-14T16:10Z). Green and red runs interleave on the same day (e.g. success 16:15 fix/issue-1133, failure 16:08 fix/detached-admission), so it is a flake, not a deterministic red.

Observations:
- The error is printed by the spawned `codex` child (agentCommand=codex, stream-json capture), in Bun's error format, resolving out of Bun's install cache (`~/.bun/install/cache`) — i.e. Bun auto-install is in play inside whatever `codex` resolves to on the runner, not the pnpm-installed workspace.
- `unist-util-visit-parents/do-not-use-color` is a conditional subpath export (node/browser) that Bun's cache resolution intermittently fails to resolve.
- The review run then reports 'All 1 file review(s) failed' and the action exits 1, failing the PR check.

Worth deciding: (a) make the action's codex invocation immune to Bun auto-install (run from an installed dir / pin PATH), or (b) treat all-file-reviews-failed as infra-neutral instead of failing the check.

roninjin10:
Closes #1546.

## Root cause — not the review action's dependencies, and not which codex binary is on PATH

`parentDeathCommand` routes every POSIX agent spawn through `parentDeathWatchdog.js`, launched as `process.execPath` — which is **Bun** whenever the engine runs under Bun (`bun runAction.ts`) — and it **inherited the agent's cwd**.

In the review action that cwd is `GITHUB_WORKSPACE`: the freshly cloned PR checkout (confirmed in the failing job log, `cwd=/home/runner/work/smithers/smithers`). That checkout contains this repo's root `bunfig.toml` with `preload = ["./preload.js"]` — and the action installs only its own tree (`pnpm -C $SMITHERS_ACTION_PATH/../../.. install`), never the caller's checkout. So the preload's import graph (`preload.js` → `@mdx-js/esbuild` → remark/unified → `unist-util-visit-parents`) has no `node_modules` to resolve from, and Bun **auto-installs it into the shared `~/.bun/install/cache`**.

Eight parallel agents (the log shows `maxConcurrency` raised to 8) then race the same cold cache entries. A lost race produces exactly the reported error, from exactly the reported path shape:

```
Cannot find module 'unist-util-visit-parents/do-not-use-color'
from '~/.bun/install/cache/unist-util-visit-parents@6.0.2@@@1/lib/index.js'
```

The watchdog dies **before codex is ever spawned**, so every file review and `narrate` fail in ~0s and the action exits 1.

This explains all three things the issue could not: the intermittency (warm vs cold cache), why #1548's PATH pin changed nothing, and why other PRs in the same batch were green.

## Evidence

1. Job log `95199278032`: the agent cwd is the workspace checkout, and the error arrives on the child's stderr at sequence 0, before any codex output.
2. Reproduced the mechanism locally — booting bun with cwd set to a directory holding that bunfig+preload and no `node_modules` auto-installs the whole graph into the cache in the exact `name@version@@@1` layout from the error string.
3. Verified Bun does **not** walk up for `bunfig.toml`, so booting the watchdog one directory away is sufficient.

**What I could not reproduce:** the resolution failure itself. Eight concurrent cold auto-installs on macOS / bun 1.4.0 all succeeded; the race appears specific to Linux / bun 1.3.13. What is made deterministic here is the *precondition* — the auto-install happening at all — and the fix removes that precondition entirely.

## Fix

The watchdog now boots in its own package directory, which carries no `bunfig.toml`, and receives the agent's cwd as `argv[2]` instead of inheriting it. `runCommandEffect` / `runRpcCommandEffect` pass `invocation.cwd` through to spawn, resolved against the engine's cwd since the watchdog no longer shares it. **The agent process still runs exactly where it was asked to.** No dependency manifests changed, so both lockfiles are untouched.

## Regression coverage

`packages/agents/tests/parent-death-watchdog-cwd.test.js` hands the watchdog a checkout whose bunfig preload throws — the deterministic form of the cache race — and asserts the watchdog still boots and the agent still runs in the requested directory. It runs under `bun test`, i.e. under Bun on Linux in CI's ubuntu shards, which is where the original failure lives.

### DIFF
--- a/apps/review/tests/action/actionManifest.test.ts
+++ b/apps/review/tests/action/actionManifest.test.ts
@@ -5,14 +5,19 @@ import { parse } from "yaml";
 type ActionStep = {
   name?: string;
   run?: string;
+  "working-directory"?: string;
 };
+
+function readSteps(): ActionStep[] {
+  const manifest = parse(readFileSync(new URL("../../action/action.yml", import.meta.url), "utf8")) as {
+    runs?: { steps?: ActionStep[] };
+  };
+  return manifest.runs?.steps ?? [];
+}
 
 describe("review action manifest", () => {
   test("pins Codex resolution to the npm-installed CLI ahead of Bun", () => {
-    const manifest = parse(readFileSync(new URL("../../action/action.yml", import.meta.url), "utf8")) as {
-      runs?: { steps?: ActionStep[] };
-    };
-    const steps = manifest.runs?.steps ?? [];
+    const steps = readSteps();
     const installIndex = steps.findIndex((step) => step.name === "Install Codex CLI");
     const reviewIndex = steps.findIndex((step) => step.name === "Authenticate and review");
     const install = steps[installIndex]?.run ?? "";
@@ -23,5 +28,20 @@ describe("review action manifest", () => {
     expect(install).toContain('codex_bin="$(npm prefix -g)/bin"');
     expect(install).toContain('"$codex_bin/codex" --version');
     expect(install).toContain('echo "$codex_bin" >> "$GITHUB_PATH"');
+  });
+
+  // #1546: Bun reads bunfig.toml from its cwd. A `bun` step started inside the
+  // caller's checkout evaluates that repo's `preload`, and because the action
+  // never installs the caller's checkout, Bun auto-installs the preload's
+  // imports out of `~/.bun/install/cache` — the intermittent
+  // `unist-util-visit-parents/do-not-use-color` failure. Every bun step must
+  // therefore stay inside the action's own installed tree.
+  test("runs every bun step inside the action's tree, never the caller's checkout", () => {
+    const bunSteps = readSteps().filter((step) => /(^|\s)bun\s/.test(step.run ?? ""));
+
+    expect(bunSteps.length).toBeGreaterThan(0);
+    for (const step of bunSteps) {
+      expect(step["working-directory"]).toStartWith("${{ github.action_path }}");
+    }
   });
 });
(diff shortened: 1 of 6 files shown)


roninjin10:
## codex-sol review: LAND WITH NITS

One finding, and it is correct:

> `apps/review/tests/action/actionManifest.test.ts:39` adds a test unrelated to this commit's production fix: reverting the watchdog changes while retaining the test still passes, because `action.yml` is unchanged from `origin/main`.

Accurate. That test asserts every bun step stays inside the action's own tree — it guards the *precondition* (a bun step running in the caller's checkout, which is what lets Bun auto-install the preload graph into the shared cache) rather than the watchdog change itself.

Keeping it, because it stops someone reintroducing that precondition in `action.yml` later, but recording plainly that it is **defense in depth, not regression coverage**. The regression coverage for this fix is `packages/agents/tests/parent-death-watchdog-cwd.test.js`, which hands the watchdog a checkout whose bunfig preload throws and fails if the fix is reverted.

Everything else came back clean: POSIX capture and RPC paths both pass the resolved agent cwd explicitly; Windows and custom-spawn paths retain cwd; relative cwd resolves against the engine; no caller, export, docs, lockfile, or repo-invariant issues.

PR #1620, opened 2026-08-17
fix(server): assert webhook metric deltas, not process-global totals

roninjin10:
Fixes the one test still failing on `test (ubuntu-latest, 3, 3)` after #1617's teardown sweep removed the `SQLITE_PROTOCOL` wedge.

## The bug

`smithers_gateway_webhooks_received_total` is an Effect `Metric.counter`. Effect's metric registry is a **process-global singleton** and its counters are cumulative. `bun test tests` runs every `packages/server` test file in one process, and CI shards by package — so every `Gateway` built in the shard shares one `{workflow="github"}` sample, and `/metrics` reports the whole process's running total, not the gateway under test.

`Gateway webhook ingestion > rejects invalid webhook signatures and records rejection metrics` asserted the **absolute** value `1`:

```js
const receivedPattern = /smithers_gateway_webhooks_received_total\{[^}]*workflow="github"[^}]*\}\s+1\b/;
```

That only holds when this file runs before every other file that posts a `github` webhook. `gateway-webhook-explicit-run-coverage.test.jsx` posts two. Bun walks the test directory in readdir order, which differs between local APFS and the runner's ext4 — locally `gateway-webhooks` happens to go first, on the runner it does not. When it doesn't, the counter is already at 2, the regex demanding `1` never matches, and the poll burns its full 10s deadline.

That is exactly the reported signature: a 10s timeout with **no SQLite error anywhere near it**, in a run where nothing stalled.

## Reproduced deterministically

Adding one test file that posts a single `github` webhook before `gateway-webhooks.test.jsx` fails it 100% of the time on the unfixed tree, and passes on the fixed one. Not a flake — a fixed order dependency.

## The fix

Baseline both counters before the request, then assert each advanced by exactly one. This matches the idiom already used elsewhere in the suite (`metricDelta` in `serve.test.js`, `metricValue` in `streamDevTools.test.tsx`) and makes the assertion independent of what else shares the process.

**No timeout was raised, no test skipped or retried, no shard rebalanced, and `jumpToFrame`'s `RewindFailed` → `Busy` rewrite is untouched.**

## Also here

`closeDb` now evicts its `SMITHERS_HOT` cache entry. #1617 made `close()` reachable, so a caller can retire a database while the hot-reload cache still points at it and hand the next import a closed connection.

## Independent confirmation of #1617's numbers

A/B on one identical tree, wrapping the `bun:sqlite` constructor, toggling only whether `closeTrackedSmithers` actually closes:

| | peak live handles | opens |
| --- | --- | --- |
| sweep disabled | **189** | 257 |
| sweep active | **53** | 257 |

Exactly the 189 → 53 over 257 opens reported on #1577.

## Gates

`packages/server` 778 pass / 0 fail · `packages/db` 837 pass · `packages/smithers` 351 pass · `pnpm lint` · `pnpm format:check` · `pnpm -r typecheck` — all clean.

(The branch was reset onto current `main` first: its previous contents were already merged as #1617, so committing on the stale tip would have reverted ~8.9k lines of `main`.)

Refs #1577

### DIFF
--- a/packages/server/tests/gateway-webhooks.test.jsx
+++ b/packages/server/tests/gateway-webhooks.test.jsx
@@ -69,6 +69,40 @@ async function postGitHubWebhook(port, workflowKey, payload, secret, deliveryId)
   });
 }
 /**
+ * Current value of one Prometheus counter sample, or 0 when it is absent.
+ *
+ * Effect's metric registry is a process-global singleton and its counters are
+ * cumulative, so every `Gateway` built in this test process shares one
+ * `smithers_gateway_webhooks_received_total{workflow="github"}` — `/metrics`
+ * reports the whole process's total, not this gateway's. Any earlier test that
+ * posted a `github` webhook (`gateway-webhook-explicit-run-coverage.test.jsx`
+ * posts two) has therefore already advanced it, so asserting an absolute `1`
+ * passes only when this file happens to run first and otherwise burns the full
+ * poll deadline and fails with no SQLite error anywhere near it — the surviving
+ * shard-3 failure in #1577. Assert the DELTA across the request under test,
+ * matching `metricDelta` in serve.test.js and `metricValue` in
+ * streamDevTools.test.tsx.
+ *
+ * @param {string} metrics raw Prometheus exposition text
+ * @param {RegExp} pattern must capture the sample value in group 1
+ * @returns {number}
+ */
+function counterValue(metrics, pattern) {
+  const match = metrics.match(pattern);
+  return match ? Number(match[1]) : 0;
+}
+/**
+ * @param {number} port
+ * @returns {Promise<string>}
+ */
+async function fetchMetrics(port) {
+  return fetch(`http://127.0.0.1:${port}/metrics`).then((res) => res.text());
+}
+const WEBHOOKS_RECEIVED_GITHUB =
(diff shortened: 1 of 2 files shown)


roninjin10:
## Shard-3 CI result: green

Run [32052900143](https://github.com/smithersai/smithers/actions/runs/32052900143) — **whole run `success`**, all 18 jobs green.

`test (ubuntu-latest, 3, 3)`:

| | before (32045785352) | this PR |
| --- | --- | --- |
| `Gateway webhook ingestion > rejects invalid webhook signatures…` | **fail**, 10065 ms | **pass**, **43 ms** |
| `(fail)` lines in the shard log | 1 | **0** |
| `locking protocol` / `SQLITE_PROTOCOL` occurrences | 0 | **0** |

The test went from burning its full 10 s poll deadline to passing in 43 ms — the counter now matches on the first poll, which is what you expect when the assertion no longer depends on another file's leftover count.

This is **one** CI run, stated plainly. Unlike the previous two attempts, though, the mechanism was reproduced deterministically before the run rather than inferred from it: forcing one prior `github` webhook post fails the unfixed test 100% of the time and passes on the fixed one. The CI run confirms a diagnosis, it isn't the diagnosis.

